PR #4, opened 2020-03-15
Bring homepage content in line with latest edits

coreycaitlin:
Hoping I didn't break anything!


aaronhans:
Thanks for the update! 

FYI We are working in the staging branch, that is the one that deploys to https://as-cdt-pub-acovid-w-p-001-staging.azurewebsites.net/ I'll merge this here and back to staging.

PR #99, opened 2020-03-23
Trailing slashes on sitemap URLS

carterm:
Our URLs have trailing slashes.  Updating Sitemap to match so search results don't contain dupes.


carterm:
Our URLs have trailing slashes.  Updating Sitemap to match so search results don't contain dupes.

PR #132, opened 2020-03-27
Steplist code

carterm:


aaronhans:
We should use our installable npm module for the javascript and have the code inlined into the footer similar to how we insert the css. This will prevent forking our published package and allow us to incorporate future updates easily

aaronhans:
Need polyfills to make this work in IE11, we can still use inlined js with the module/nomodule pattern just for modern browsers and have the larger polyfilled bundle loaded at an external.js url so we don't slow down all pages for legacy browsers

PR #149, opened 2020-03-30
Clean output folder

carterm:


carterm:
Nevermind.  It's got all sorts of random stuff from staging in it.

PR #190, opened 2020-04-05
fix bug with button submit

aaronhans:
counties not being passed into template from form submission via button click (only worked via autocomplete submit)


abquirarte:
thank you @aaronhans !

PR #247, opened 2020-04-13
README formatting

tallguyjenks:
- capitalize title
- separate text from headings so the spacing follows convention
- indent links for aesthetics?


abquirarte:
thanks @tallguyjenks

PR #253, opened 2020-04-13
fix telehealth

aaronhans:


aaronhans:
The WP post is now a draft and referenced as an include instead of a tagged collection which is simpler

PR #259, opened 2020-04-14
Homepage quick wins

artuoma:


aaronhans:
I thought we weren't moving this up unless we got faster stats updates

abquirarte:
It's a temporary move it. I had already requested this to be moved up. 
We can move up even more once we think through new design and automate whatever we can for accurate/faster results @aaronhans :)

PR #277, opened 2020-04-15
Fraggle

aaronhans:


aaronhans:
review this first to simplify this diff: https://github.com/cagov/covid19/pull/280/files?diff=split&w=1

aaronhans:
This makes more sense now, thanks for reviewing the format change separately @artuoma

PR #279, opened 2020-04-15
Format

aaronhans:


artuoma:
This is a bunch of changes, are you trying to use more partials or what's the goal?

PR #280, opened 2020-04-15
format main template

aaronhans:


aaronhans:
review this PR before https://github.com/cagov/covid19/pull/277 then the diff in fraggle will be simpler

PR #284, opened 2020-04-15
Adding daily NEW! section as a WordPress partial

artuoma:


aaronhans:
Can we review and merge this first because it adds more partials to the same template: https://github.com/cagov/covid19/pull/277

PR #287, opened 2020-04-15
Fixing pagenav for more lang support

carterm:
ready to add more languages.  Or at least the last structure has been upgraded to be scalable.


carterm:
in staging.

Issue #301, opened 2020-04-16
Improve Accordion editing process


artuoma:
- [x] Update accordions on resources for emotional support and well-being
- [x] Update accordions on Stay home order
- [x] Update code to support new "wp-accordion" class
- [x] Simplify WordPress editing process (Thanks @carterm !)
- [x] Update WordPress pages that have accordions
- [x] Update documentation

artuoma:
Currently, only one content type block is supported directly under the title. To use multiple content blocks such as a list and a paragraph, those elements must be grouped together. Additionally, a certain CSS class must be applied. 

We are currently working on a solution to negate this complicated process.

artuoma:
Working with @aaronhans and @carterm to simplify editing further.

carterm:
https://github.com/cagov/covid19/pull/307

artuoma:
[Documentation updated](https://covid19.ca.gov/stay-home-except-for-essential-needs/#top) and WP pages updated. Thanks all!

carterm:
Done with the old accordion code yet?  I'd like to remove it.

PR #307, opened 2020-04-17
Proposed find accordions

carterm:


carterm:
> excellent @carterm
> 
> Can we move this out to another file and include it because the 11ty config is getting long

let's make a "cleanup the 11ty config" issue.

carterm:
https://github.com/cagov/covid19/issues/301

PR #335, opened 2020-04-22
accordion removing "wp" class from header

carterm:
forgot to remove the wp only class after processing the header


carterm:
https://github.com/cagov/covid19/issues/301

PR #346, opened 2020-04-23
Data drive web.config redirects

carterm:
Using 11ty to generate redirects in the web.config from Wordpress data.


carterm:
https://github.com/cagov/wordpress-integration-api/issues/16

PR #376, opened 2020-04-27
Spiffytele

aaronhans:
The telehealth lookup form now uses our published web components that handles the form and autocomplete against the specified endpoint


aaronhans:
Using new published component: https://www.npmjs.com/package/@cagov/lookup

artuoma:
@carterm should check this out.

PR #377, opened 2020-04-27
accessibility fix

aaronhans:
missed this in weekend update, need a11y label for button with no text


aaronhans:
this is obsolete now

Issue #409, opened 2020-05-01
Remove timestamp from search page 


abquirarte:
The time stamp on search confuses users. It makes it seem like the website has not been updated when they search. 

Can we remove it? 

![Screenshot_20200501-124830_Chrome](https://user-images.githubusercontent.com/9556188/80836688-22a28980-8baa-11ea-90df-f4e7c8cde58e.jpg)

aaronhans:
https://github.com/cagov/covid19/pull/412

PR #412, opened 2020-05-01
Langpulldown

### LINKED ISSUE #411: fix multi county zips
A zip code: 94505 was totally missing from our dataset

This zip code covers parts of both Contra Costa and San Joaquin counties so we need the service and frontend to support returning multiple county matches for a single zip code

aaronhans:
These changes go live with the Chinese simplified and traditional site translations
also fix #411 and removes date from search page


aaronhans:
fixed #409 too

PR #435, opened 2020-05-07
handle phone number fields with text in them

aaronhans:


aaronhans:
have to merge this or we will be linking the text "see insurance card" which is now used in place of phone numbers some places now @abquirarte

PR #467, opened 2020-05-12
delete the telehealth manual page

aaronhans:
We don't want to use a manual page for telehealth anymore, we will use a normal WP post with a custom HTML block for the search widget. The new post is already created, just need to delete this manual file


aaronhans:
I put a redirect in the redirect table from /telehealth to the new version that has everything on it published from wordpress currently at /telehealth-care we can reclaim the old url when this is deleted

PR #471, opened 2020-05-12
adding stages share image to all roadmap pages

artuoma:


aaronhans:
this update is in staging now

PR #534, opened 2020-05-18
Website redesign go live Monday

artuoma:
- Homepage header structure and style changes
- Homepage layout
- Colors and styles


pages/manual-content/index.html
```diff
@@ -1,6 +1,6 @@
 ---
 layout: main.njk
-title: California Coronavirus COVID-19 Response
+title: COVID19.CA.GOV
```
aaronhans, line 3: Did you want this as the title?

pages/manual-content/search.html
```diff
@@ -21,6 +24,8 @@ <h2>Search this website</h2>
 	<button type="submit"  class="btn btn-primary">Search</button>	  
 </form>
 
+<div id="answersNow"></div>
```
aaronhans, line 27: I don't think we want to release the Q&A stuff yet, this will make it live

pages/manual-content/index.html
abquirarte: Will this new title affect SEO?

Main header: California Coronavirus COVID-19 Response
H1:  California Coronavirus COVID-19 Response

Title Current:  <title> Coronavirus COVID-19 Response Portal</title>

Notes: The title will show on google search results I would add State of California  to the main title.

pages/manual-content/search.html
abquirarte: true, not sure we are ready for this. Unless we are ready to collect and act on all the feedback... hahaha.

artuoma: Q&A just goes to the Stay home page

pages/manual-content/index.html
artuoma: Seems like the GOV was ok with it and it doesn't need to be translated since it is a URL

artuoma: @abquirarte

pages/manual-content/search.html
artuoma: @abquirarte

pages/manual-content/index.html
abquirarte: Mhm. I worry about SEO after this. Let's go for it but stay open minded.... cc @carterm for awareness.

pages/manual-content/search.html
abquirarte: so it it still brings the new QA only on the stay home page

artuoma: Check the live site and see if that is what you want.

aaronhans: This is going live with the quick answers on search results. We can go for it now and respond to feedback but we planned to user test and review our quick answer database for a few more days. I don't know if you have any intentional changes on this page or not.

PR #536, opened 2020-05-18
fix missing announcements on translated pages

aaronhans:


aaronhans:
the plasma.js is a generated file which shouldn't be committed

carterm:
This PR broke the stats for translated pages. @aaronhans

PR #543, opened 2020-05-19
missing fallback check

aaronhans:


aaronhans:
This fixes missing header image on safari

Issue #556, opened 2020-05-19
dynamic case data


aaronhans:
High level goals:

- [x] Build API that retrieves data from snowflake
- [x] Make stats on homepage powered by API
- [x] Create first D3 version of clickable county map, compare weight vs tableau

Task limits:

We don't plan to spend much time experimenting with different data visualizations but we can use the version the GO is happy with as an example and build a faster, more web safe version that is easily embedded, doesn't risk the tableau bug where the numbers disappear at different screen widths

The source of some data points like PPE distribution is still a mystery, going to leave that out of v1

abquirarte:
This will be great, thank you!

aaronhans:
More action items, when confident API powered stats will continue to work:

- [ ] Delete the table-data file in WP
- [ ] Delete the unused stats-lang files in translated-posts unless those are better versions, then replace the WP versions with those...

PR #579, opened 2020-05-20
get stats from data

aaronhans:
Get the stats data from the file the API is writing instead of the WP table-data

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -4,7 +4,7 @@ const { JSDOM } = jsdom;
 const fs = require('fs')
 const langData = JSON.parse(fs.readFileSync('pages/_data/langData.json','utf8'));
 const pageNav = JSON.parse(fs.readFileSync('pages/_data/pageNav.json','utf8'));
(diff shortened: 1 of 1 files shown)


carterm:
So are we done using the Wordpress table for this then?

aaronhans:
That is the plan. I need to make sure I understood the snowflake table structure properly. The first location I used to get this data went away. I think I am using a stable set of tables now.

I added tasks to cleanup the unused files to https://github.com/cagov/covid19/issues/556

abquirarte:
So is this operational now? Want to make sure we are cross-validating with
what CDT has on the update page just in case this goes through other data
curation that we are not aware of.

<https://twitter.com/cadotgov>

aaronhans:
This is operational now. Karim is no longer making manual updates to the stats data. @abquirarte Is there a dba or someone who can advise me on the best place to be retrieving different pieces of information? I have manual and programmatic SQL access but some of the tables are fluid.

Issue #595, opened 2020-05-21
create image optimization pipeline


aaronhans:
- We are starting to deal with more images in pages
- We often get many different files for a single placement because we have localized versions
- We need to create multiple output file types in order to take advantage of the best available compression in each browser

We can use squoosh.app to handle this manually but that service was down this afternoon

Solution:
- [x] Write optimization scripts to compress source files into multiple output formats

PR #596, opened 2020-05-21
image optimization scripts and first batch

aaronhans:


aaronhans:
https://github.com/cagov/covid19/issues/595

PR #605, opened 2020-05-22
localized images, changing these now, received translation should be …

aaronhans:
…identical

also made took several seconds off the build because the css minification was running 400 times at 22ms but could be cached the first time


abquirarte:
THANK YOU 🙏

PR #609, opened 2020-05-22
faster builds by tracking hashmap of previously transformed html

aaronhans:
This optimization takes 75% off processing time for builds running locally. It won't have any effect on the builds on github server because the stored but ignored local hashmap file will have to be rebuilt in ephemeral build container but the main annoyance of slow builds was dealing with them locally


aaronhans:
I reverted the pageNav update that hid contact tracing in here too since it is now ready for release

PR #696, opened 2020-06-02
can pass today to have it display current date

aaronhans:


aaronhans:
The plan is to display the current data always on roadmap-counties because it has the tableau embed which is constantly updated in it...

Could be a better way to do this, please advise if one occurs to you

PR #709, opened 2020-06-04
Accessibility fixes june

aaronhans:
all changes from Art


aaronhans:
changes look good, runs fine locally

PR #710, opened 2020-06-04
Adding a redirect for any content that is in the "manual-content" folder

carterm:
This will remove "manual-content" from any url by redirect.


carterm:
(Tested on my azure instance)

PR #748, opened 2020-06-09
search add header

aaronhans:
style modification to search results: adding header above the google site search results and hiding the padding and relevance drop down


abquirarte:
✔️

PR #769, opened 2020-06-11
replacing-great-plates-participant-faqs

ashleydraper:
expiration dates still reads June 10th


aaronhans:
It looks like this filename is missing the .pdf extension

aaronhans:
This live file https://covid19.ca.gov/pdf/great-plates-delivered-participants-faqs.pdf shows July 10 as the program expiration

aaronhans:
We got this sorted out, updated the link in WP and added a redirect from the old wordpress uploaded pdf url to the new github uploaded pdf url

PR #816, opened 2020-06-16
adding translation pages for translation page

carterm:
The text for translation page should be translated


carterm:
Enable WP page when deploying this.

Issue #535, opened 2020-05-18
Add COVID-19 hotline number to the footer


thepegisin:
- Remove "Statewide COVD-19 Hotline" from footer.
- On new line, put "COVID-19 hotline: 1-833-422-4255 M-F 8AM-8PM, Sa-Su 8AM-5PM" 
- Make hotline number a dialable link
- Link [https://covid19.ca.gov/hotline/]((https://covid19.ca.gov/hotline/)) to **Hotline information** menu item in new IA and rename that **COVID-19 hotlines**

TamaraIDEJA:
This is GREAT! Thank you @thepegisin for trying to make hotline info more prominent!

abquirarte:
@artuoma can you do this? or Konstantin? 
I would try it, but don't want to break the site.

artuoma:
Yeah we can have @koryfey try to see if there is any issues working within the established process.

abquirarte:
perfect :) I thought about tagging him.

Angelica Quirarte, "Angie"
Assistant Secretary- Government Operations Agency
c: 916-526-5510
Digital Crisis Standard- covid19.ca.gov
<https://github.com/cagov/covid19/wiki/Crisis-Standard>

Follow us:
Web <https://alpha.ca.gov/> | Blog <https://medium.com/cadotgov> | Twitter
<https://twitter.com/cadotgov>

artuoma:
@kkoryfan

kkoryaka:
Added changes to staging. Please review.

abquirarte:
Am I looking at the wrong staging? 
https://as-cdt-pub-acovid-w-p-001-staging.azurewebsites.net/

abquirarte:
Thank you so much @kkoryaka !

abquirarte:
I may have closed it too early. Staging looks good. We should push to production @kkoryaka 

CC: @artuoma

kkoryaka:
I have made the pull request to the master. Now we have to wait when someone reviews it and make it live.

thepegisin:
It's live and looks great. Thanks, @kkoryaka! Closing this now.

abquirarte:
Yay!!
Angelica Quirarte, "Angie"
Assistant Secretary- Government Operations Agency
c: 916-526-5510
Digital Crisis Standard- covid19.ca.gov
<https://github.com/cagov/covid19/wiki/Crisis-Standard>

Follow us:
[...]

PR #821, opened 2020-06-17
Added hotline number in the footer per issue #535

kkoryaka:
Added new line with hotline info to footer.njk


abquirarte:
☝️ interesting behavior (new from my lens) @aaronhans

PR #829, opened 2020-06-17
turning off sitemap for search translated pages

carterm:
Translated pages need to inherit the tags of their original pages.  This is a temporary fix to seach pages not being removed from sitemap.


carterm:
Created this issue for resolution...
https://github.com/cagov/wordpress-integration-api/issues/36

PR #892, opened 2020-06-23
new translation filters 

aaronhans:
logic is:
- if we are in a translated file
- look for a string pattern in image sources and links
- if we find it see if there are corresponding translated files on our production CDN already
- if translated versions exist rewrite the link in the post to use the local file

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -3,6 +3,7 @@ const jsdom = require("jsdom");
 const { JSDOM } = jsdom;
 const fs = require('fs');
 const md5 = require('md5');
+const fileChecker = require ("https");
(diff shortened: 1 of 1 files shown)


carterm:
I feel like rather than checking the storage for the file during the build, we should just assume that any --en postfix has at least placeholder files and simply upload English copies until the translated versions are ready.  I'm concerned that there could be a failed response during the build that will result in links being written for English.

carterm:
We can also enable "list blobs" permissions on the storage and return a manifest of existing blobs, rather than a bunch of HEAD requests.

carterm:
...but...this is a great solution for now.

Issue #794, opened 2020-06-15
better pdf delivery phase 1


aaronhans:
We can meet several of our requirements immediately with a separate pdf repository that deploys to blob storage backed cdn and is served by a redirect at covid19.ca.gov/pdf

The benefits we will get:
- PDFs can be deployed to production much faster
- Main site build size will shrink so its build will speedup slightly
- Better resilience because all pdf changes happen in different pipeline from site and cannot break it

Tasks
- [x] Create new repo (@aaronhans)
- [x] Setup new blob storage, CDN (@ChadABratton, Shamal)
- [x] Hook up new deploy pipeline (@ChadABratton)
- [x] Set CDN so that it has < 1 minute cache time for pdf files (@ChadABratton)
- [x] Create test redirect using web.config in Windows env (@carterm)
- [x] Test all links because this will be case sensitive, fix all broken ones (@aaronhans)
- [x] Fork, upgrade azure upload git action to be more efficient by only pushing changed files (@aaronhans )
- [x] Setup friendly domain name for new CDN (@ChadABratton)
- [x] Use in production

aaronhans:
Docs on how to manage files in subdirectores and latest domain info added to readme: https://github.com/cagov/covid-static

carterm:
This is done. @aaronhans @ChadABratton .

PR #909, opened 2020-06-29
Nopdf

aaronhans:
This deletes the entire pdf from production

- We deployed a redirect on Thursday which sent all requests to /pdf to files.covid19.ca.gov
- This will slim down the build artifact significantly


aaronhans:
related to https://github.com/cagov/covid19/issues/794

carterm:
Whoa.  Scary.

PR #937, opened 2020-07-05
revert stats to fix conflict

aaronhans:


abquirarte:
This is the fix to the other survey?

PR #982, opened 2020-07-13
add survey in spanish

aaronhans:
This adds survey url to data parameters
survey prompt translation has been added to the footer data table in WordPress
reconfigures the survey code to be a web component
both the survey language prompt and the url come from site data now and are language dependent


aaronhans:
This is not the correct url, waiting for approved url from Hilary

PR #1133, opened 2020-07-29
search rounded

carterm:
Needs some styling for left align when typing.


carterm:
Closing as part of a larger design.

PR #1135, opened 2020-07-30
Cross-platform dev support with Windows

xjensen:
This change adds cross-platform support for developers of the site. No code changes will be necessary when switching between Mac, Linux, and Windows. 

It's a small change, but I didn't want to commit directly to `trilogydev` and risk breaking everyone's local builds. Hence, a pull request.

Note that this change brings in a new devDependency, [shx](https://github.com/shelljs/shx). You'll probably need to `npm i` before this will work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4072,6 +4072,12 @@
         "is-symbol": "^1.0.2"
       }
     },
+    "es6-object-assign": {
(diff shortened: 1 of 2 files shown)


abquirarte:
Cool! Thank you!

PR #1142, opened 2020-07-30
fix var in news loop

aaronhans:


carterm:
My bad.  Noobie mistake.

PR #1197, opened 2020-08-04
HTTP => HTTPS, covid19 => files.covid19 for PDFs

carterm:
This PR will only be for fixing host link parts.


carterm:
Going to scrap this PR because we don't need to fix all this right now.

PR #1198, opened 2020-08-04
Manual translations without https and files host

carterm:
updating links for specific pages


carterm:
This has been pushed to staging.

PR #1223, opened 2020-08-06
Removing search results page from search results

carterm:
Translated search results pages should not be indexed.

Need to investigate how this got reverted,


carterm:
This was caused by this issue...
https://github.com/cagov/wordpress-integration-api/issues/36

Issue #801, opened 2020-06-15
Windows compatability


aaronhans:
Thanks for discovering how to get the covid site build working on Windows Mark!

Is there any way to get that build script to work cross platform? Let's do some research on how other people have handled this and see if there is any way we can make that happen even if we have to slightly change how our scripts work by using code to to the copy operation or something.

We should immediately add the directions you discovered to get it working on Windows to the readme.md

Mark-Tyler:
Here is what the differences:

On Windows

line 47 of package.json    "build": "npm run build:rollup && npm run build:ancient && webpack --mode production && copy .\\docs\\css\\build\\built.css .\\pages\\_includes\\built.css",

On Mac

line 47 of package.json    "build": "npm run build:rollup && npm run build:ancient && webpack --mode production && cp ./docs/css/build/built.css ./pages/_includes/built.css",

aaronhans:
Thanks Mark! I think we can figure out how to do that copy differently so it works cross platform. The css file is created by webpack and needs to be in place for the 11ty build. We may be able to copy it at the beginning of the 11ty build script or we can use some other cross platform copy script to handle it at the right order

aaronhans:
thanks @xjensen! https://github.com/cagov/covid19/pull/1135#pullrequestreview-458035415

xjensen:
Quick update. While my earlier quick fixes did get the site to build on Windows, I continued to see a lot of problems in the local development watch/serve functions. So I kept working. That work can be found in the `trilogydevForWindows` branch. Here's a list of the high-level changes.

* Webpack has been removed from the build system. 
* For JS builds, everything now runs through Rollup. 
     * Babel is now integrated into the es5.js Rollup file, `rollup.config.es5.js`. This includes the preset-env module and browserlist settings. This mimics the old Webpack set-up for this file.
     * There's a combined Rollup file that can process all JS files in one pass, `rollup.config.es5.js`. Individual Rollup configs still live in their respective folders.
     * JS builds are now triggered through Gulp.
* For CSS builds, everything now runs through Gulp.
     * Instead of Webpack, there's a Gulp task to call node-sass directly. 
     * After SCSS is built, CSS purging and minifying is performed through PostCSS. Adding additional PostCSS modules here, such as autoprefixer, would be easy.
     * Because minifying is now performed during the build, we no longer need to do so within an Eleventy template filter. Those functions have been removed from the templates and `.eleventy.js` config.
[...]

PR #1250, opened 2020-08-10
Rebuilding the build system

xjensen:
The build system gave me a lot of trouble on Windows. So I rebuilt it. It's working consistently across both Mac and Windows. Here are my testing platforms at the moment.

* Windows 10 with Node 12.18.3.
* MacOS 10.15 (Catalina) with Node 14.7.0.

**Much has changed.** Here's a list of the highlights.

* Webpack configs have been removed.
* Additional Rollup configs have been added.
     * The es5.js file is now generated through Rollup instead of Webpack. See *src/js/rollup.config.es5.js*.
     * There's a new Rollup config that bundles together all other Rollup configs, so they can all be called at once. It's faster. See *src/js/rollup.config.all.js*.
* Gulp has been added to help manage build steps. See *gulpfile.js*.
* CSS builds are now handled through Gulp instead of Webpack.
* CSS minifying is now handled during the build process, rather than through Eleventy's template filters. Related Eleventy code has been removed from *.eleventy.js*.
* Separate watchers for webpack/rollup/eleventy have been replaced with Gulp-based watchers and a custom implementation of Browsersync. 
* There's a new "development mode" for rapidly prototyping CSS changes. The goal is to enable rapid iteration on CSS without requiring costly Eleventy and PurgeCSS rebuilds for every change.
     * There's a new *pages/_data/env.js* file which exposes environment data to Eleventy templates.
     * Note small but meaningful changes to how CSS is included in *pages/_includes/main.njk* and *page.njk*.
     * In this mode, CSS is not purged and embedded into the HTML. Instead, a full, un-purged CSS file is used across the site. 
* Npm scripts have been modified to accommodate these changes. 
     * `npm run build` builds everything.
     * `npm run build:css`, `npm run build:js`, and `npm run build:site` perform respective functions.
     * `npm run clean` will wipe out the *docs* folder if you want a fresh start.
     * `npm run deploy` will do a `clean` prior to a `build`.
     * `npm run local` will build and serve the site for live local development.
     * `npm run local:dev` will enable the *development mode* of `local`, see above.

Further information about rationale for these changes can be found in #801.

Changes here are pretty extensive, so I'd say plenty of tire kicking is in order before proceeding with the merge.


xjensen:
Here's a list of outstanding questions.  

* The current Webpack-based CSS builds create some files (in the *dist* folder, if memory serves), but I'm not sure if they're needed. Everything seemed to be covered already by Eleventy's Passthrough Copy. There are some other hash-named JS files going to *docs/js* which do not seem necessary. Still, I'm not sure if a downstream build system needs this output, etc. It would be good to get a second opinion.
* I've tested the new legacy ES5 Javascript in IE11. It seems to be working as well as the old configuration, at least. If there are any other key browsers among our users we need to check, I'll be happy to take a look.
* It now occurs to me that the new CSS development mode could easily be extended to Javascript embeds as well. We can quickly add those changes if desired. Or remove development mode altogether.

xjensen:
I just realized I still need to extract static assets from imported web components. I'll do that next.

carterm:
Oh wow this is great!  I was able to run it locally.

aaronhans:
This is incredible work @xjensen amazed you have been able to fit all this progress into nights and weekends. Thank you for all these improvements.

xjensen:
Thanks!

I've chunked out some of these changes into smaller PRs: #1284, #1285, and #1286. These new PRs do not cover everything here. They're a second step following the work by @carterm earlier today on #1276 and #1283.

aaronhans:
Having a separate dev config where the css isn't inlined is a big improvement. I wonder how costly the purge step is because I  am a little nervous about people getting used to working in dev mode and pushing updates without verifying everything is cool after the purge is applied (This is probably a bug that needs to be caught visually)

xjensen:
In my testing so far, the CSS purging in this new Gulp-based build system takes two or three seconds. (That's two or three seconds to purge the home.css and built.css files in parallel.) It's slower than instant, but still not bad. We could definitely give it a shot.

PR #1277, opened 2020-08-12
Dependabotprs

aaronhans:
merging all the open dev tool dependency upgrades and telling dependabot to open PRs against this branch in the future


aaronhans:
These changes have already been deployed to staging, and the github action run still worked

PR #1284, opened 2020-08-13
Combining Rollups into one file

xjensen:
Adds a new Rollup file (_src/js/rollup.config.all.js_), which references all the other Rollup files. Using this combined file speeds up the build process by removing a lot of execution boilerplate. Only one single Rollup call is needed, which means fewer hops through `npm run`, etc. 

None of the existing Rollup files are modified in any way. You can continue using them as-is. This should be a fairly safe change. 

This change is an incremental chunk of the larger PR #1250. Unlike #1250, here I'm **not** including the new ES5 config within this all-Rollup file (see related PR #1285). This means you could put this new file into service within the existing build system, no fuss. When ready, you could swap `npm run build:rollup` with the new command here, `npm run new:build:js:all`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "watch": "onchange 'src/**' -- npm run rebuild",
     "rebuild": "npm run build && touch pages/_includes/main.njk",
     "local": "npm run build && run-p watch start",
-    "compress:img": "cd src/build && node img.js"
+    "compress:img": "cd src/build && node img.js",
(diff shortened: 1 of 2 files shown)


abquirarte:
😍

carterm:
@xjensen Can you update this to just replace the existing "build" script so that the pipelines recognize it?

aaronhans:

PR #1285, opened 2020-08-13
Rollup for ES5

xjensen:
Adds a new Rollup file (_src/js/rollup.config.es5.js_) to process legacy ES5 Javascript for older browsers. The intention is to eventually replace the current Webpack file (_src/js/webpack.config.es5.js_) which does the same. Phasing out Webpack would allow us to run all JS through Rollup, a significant simplification within the build process. Compared to the Webpack file, this new Rollup file still uses a very similar Babel configuration. 

This should be a relatively safe change as no existing functionality is touched. Yet. I'm assuming we'll want to test this new file more thoroughly before putting it into real service, to ensure nothing breaks. That prospective testing can be done here via the new `npm run new:build:js:ancient` without breaking the current site. I've done preliminary testing in IE11 to my satisfaction. But I don't know where the gremlins hide, and I'd be keen to learn if we need to check any other legacy browsers.

This change is an incremental chunk of the larger PR #1250.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1222,6 +1222,31 @@
         "fastq": "^1.6.0"
       }
     },
+    "@rollup/plugin-babel": {
+      "version": "5.1.0",
+      "resolved": "https://registry.npmjs.org/@rollup/plugin-babel/-/plugin-babel-5.1.0.tgz",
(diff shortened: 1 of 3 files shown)


aaronhans:

PR #1290, opened 2020-08-13
Gulp-based build system

xjensen:
Here we have another incremental change from the larger PR #1250. See related previous merged PRs #1276 and #1283 (by @carterm) and my new PRs from last night: #1284, #1285, and #1286. 

This is the first crack at implementing Gulp as part of the new build system. Current build processes rely upon Unix-style commands like `cp` and `touch`. This works great on Mac but falls apart on Windows. Also, the current combination of multiple watchers (webpack, rollup, onchange, eleventy, etc.) makes it difficult to trigger refreshes at the correct/efficient moments during local development. Again, this is kinda okay on Mac. But Windows really doesn't like it. Gulp gives us the tools we need to create a native Javascript solution, to solve these issues in a cross-platform way. 

So instead, we're using Gulp's watchers to handle all building. We're then implementing our own configuration of browsersync, so we can trigger local reloads at more opportune moments during the build process. You have a bunch of new commands to play with.

* `npm run new:gulp:build` - builds the site
* `npm run new:gulp:build:site` - builds eleventy
* `npm run new:gulp:build:css` - builds the CSS through Gulp
* `npm run new:gulp:build:js` - builds the JS through Rollup
* `npm run new:gulp:clean` - deletes the contents of the _docs_ directory
* `npm run new:gulp:deploy` - performs a `clean` before doing a `build`
* `npm run new:gulp:local` - fires up the watchers for local development

Here's what this PR does **not** do, compared to #1250. 

* The _development mode_ described in PR #1286 is not yet enabled here. #1286 will need to be merged before we can realize that work.
* The legacy ES5 Javascript file is not yet produced by this build system. The corresponding ES5 Rollup file can be found in #1285. I'm waiting on comment and/or merge on that PR before proceeding here. I'd rather not do the work to implement Webpack into Gulp if we're going to merge #1285 soon.
* Likewise, we can speed up JS processing by merging #1284 and implementing that here.

While there's much to chew, these changes leave all of the current build process in place. You can play with this separately. I don't think this is ready for prime-time until we work through some of the other PRs mentioned above. That said, I believe it's safe to merge this in to allow further development on the side.


aaronhans:
Thanks for summarizing the reasoning here. The introduction of gulp was the bit I wasn't sure about but this sounds reasonable

aaronhans:

Issue #1299, opened 2020-08-13
make the build system hella faster


aaronhans:
faster please

aaronhans:
I am going to link all the PRs you've got going here @xjensen 
I'll merge them to staging, then trilogydev and do some testing

aaronhans:
split up PRs:
https://github.com/cagov/covid19/pull/1290
https://github.com/cagov/covid19/pull/1286
https://github.com/cagov/covid19/pull/1285
https://github.com/cagov/covid19/pull/1284

big one being split into smaller ones https://github.com/cagov/covid19/pull/1250

xjensen:
Thanks, @aaronhans! Feel free to ask any questions. 

We'll likely need additional commits and/or PRs to bring these four smaller PRs together into something resembling #1250. Next steps will depend upon what makes the cut from this first round. For example, I don't believe any of my smaller PRs include the `npm` script needed to trigger the development mode. I can add that later, for sure; it just doesn't make sense as a small stand-alone PR until a few of these other PRs are merged.

If there's anything I can do to help further, just let me know.

aaronhans:
I've got stuff merged into trilogydev and staging. I replaced the gulp build step in the trilogydev branch because we are doing a lot of work in there.

xjensen:
Two new PRs related to this: #1305 and #1306. Both are pretty basic. I think the `trilogydev` branch may be going without an _es5.js_ file at the moment, so #1305 (or a different Rollup/Webpack combo) may be important there.

New CSS builds are currently failing for me on the `staging` branch due to duplicate basenames between the _cagov-font.css_ and _cagov-font.scss_ files. Deleting one file or the other seems to be making different icons fail. Interesting. But I'm wondering if I should spend the time to troubleshoot since `trilogydev` will eventually solve this problem anyway.

One idea. I'm thinking we could speed up the Eleventy build process by moving some of the heavy HTML manipulation upstream, to Wordpress or some other process. I'm specifically thinking about the function that converts the Wordpress accordion into a web component. Would writing a Wordpress plug-in be off the table for any reason? I don't know much about how that part of the site works so I'm just making guesses about what's possible.

aaronhans:
I was thinking the same thing about the 11ty build process: https://github.com/cagov/covid19/issues/1154

aaronhans:
@xjensen can we try something like this as a strategy to get all this merged:
- open PRs from master but target trilogydev
[...]

PR #1305, opened 2020-08-14
Add es5 to all rollup

xjensen:
A couple simple changes to add the ES5 Rollup config into the all-Rollups config file. This follows previous merges related to #1299.


xjensen:
You may want to give this one special attention with respect to the `trilogydev` branch. I think the current version of the all-Rollup file there does not include the ES5 config, which I think means no es5.js file is getting generated on that branch at the moment.

PR #1395, opened 2020-08-24
New build system

aaronhans:
Putting @xjensen's new build system into production


xjensen:
Currently porting over recent additions from `trilogydev` and doing a final review.

PR #1414, opened 2020-08-25
Fixing font waffles

carterm:
Fixing font root path to work on subpages.


carterm:
Mistake, meant to merge to dev.

Issue #1294, opened 2020-08-13
homepage design: make video section data driven


aaronhans:
This got hardcoded in our trilogydev branch. We need to make it data driven again so it gets translated properly and can be updated by the content team

carterm:
This is mostly structure, so a bad candidate for a Wordpress file.  I'll work on making the English headers extractable.

aaronhans:
This might be a little tricky: We are going to want to let the content team swap in new videos at will @carterm 

Also they are going to probably need to use a different video on the translated homepages

carterm:
Added a "VideoID" that can be changed and set per language with all the other translations in the homepages.
https://github.com/cagov/covid19/pull/1454

PR #1454, opened 2020-08-27
Added VideoID by language

carterm:
VideoID added by language to the main pages.

(Lets you set a different video per language)


carterm:
https://github.com/cagov/covid19/issues/1294

PR #1509, opened 2020-08-31
Added text wrap to translations logo title in mobile.

kkoryaka:
Also fixed some padding to find the county status button


aaronhans:
@koryfey the classname update in the HTML file has to be applied in WordPress

aaronhans:
So we should remove that file from this PR, then the conflict will resolve

PR #1510, opened 2020-08-31
CA For All home links translated

carterm:
Both CA For All links will maintain language now.

(Still need to fix mobile home link)


carterm:
> I like the conditional use of `set` here.

Yah, I like that better too.

PR #1514, opened 2020-08-31
updated tuolomne

aaronhans:
The new dataset has been deployed to staging: https://staging.covid19.ca.gov/safer-economy/


aaronhans:
I think we should deploy this but the map below our activity section is not yet updates, maybe we should synchronize deploys... @hannahslee

PR #1540, opened 2020-09-01
add data and tools to menus

aaronhans:


abquirarte:
Thank you! Will await on review from @xjensen and @carterm 😄

PR #1544, opened 2020-09-02
highlight template literals

aaronhans:
Using comment syntax to trigger javascript template literal syntax highlighting in VSCode using this plugin: https://marketplace.visualstudio.com/items?itemName=Tobermory.es6-string-html


aaronhans:
This works nicely but if you have other methods for doing syntax highlighting I'm happy to use those. If we were using tagged template literals it would highlight without the comment but there isn't any universal processing required in many cases.

PR #1575, opened 2020-09-03
Javascript in Dev Mode

xjensen:
I've been doing a lot of Javascript lately, so I decided to speed up JS while working in development mode. Here are the highlights.

* The es5.js file is not built when running in dev mode. This was taking a majority of the time in the JS build cycle. Seemed prudent to take it out. It's still included in normal builds and `npm run start`.
* In dev mode, JS is not embedded. Instead, it's called from a file.
* Rollups have been revised to choose different output paths between dev and full modes.
* Created Nunjucks macros to conditionally include JS/CSS based on environment.
* Edited Gulp to reload JS without an Eleventy build in dev mode.

As with CSS, we shouldn't expect any changes to the final output when *not* running in dev mode. Behavior should only change when doing dev.


pages/_includes/main.njk
```diff
@@ -1,3 +1,5 @@
+{% import "macros.njk" as macros %}
```
aaronhans, line 1: cool

src/js/rollup.config.all.js
```diff
@@ -3,20 +3,19 @@ import es5 from './rollup.config.es5';
 import esm from './rollup.config';
 import plasma from './plasma/rollup.config';
 import reopening from './roadmap/rollup.config';
-import survey from './survey/rollup.config';
```
aaronhans, line 6: why don't we need survey anymore?

aaronhans:
Great plan!

pages/_includes/main.njk
xjensen: [Nunjucks feature](https://mozilla.github.io/nunjucks/templating.html#macro), seemed useful here.

aaronhans (approved):

PR #1588, opened 2020-09-04
new homepage and schools language

aaronhans:
This update includes:

- Changing h1 text on the homepage to labor day specific language for most languages. Tagalog and Vietnamese are not included because the translated text was so long it broke the layout on some mobile phone sizes
- Special logic for schools because they do not conform to regular tier guidance which is quite weird. This is supposed to be temporary

These updates are browsable in staging now


abquirarte:
🎉 🍻 
Thank you @aaronhans !!!!! 🦸

PR #1609, opened 2020-09-09
remove messed up press release

aaronhans:
delete this badly scraped press release


aaronhans:
Also:
- Remove few more press releases that were not related to covid19 but got scraped because there was covid in a url somewhere inside the summary body
- Rewrite the news template to use a template literal. This page gets flagged for slowing down the build because it is rebuilt in each language and is very long

PR #1647, opened 2020-09-12
only build most of the js bundles when the files in those dirs changed after the last generated file

aaronhans:
Compare file modified dates of source files to generated files before running rollup bundling.

Why do this?
- The rollup bundling is taking around half a second per separate bundle if there are any imports
- This change skips any bundle whose source hasn't been updated more recently that its generated output
- This takes a few seconds off builds with js changes during development

This seems weird what are other possible ways to handle this?
- Usually this is handled in one big bundle with dynamic imports

Why don't we put everything in one js file and use dynamic imports like everybody else?
- This is faster for the end user
  - If we had one big bundle the page specific js would not be requested until the dynamic import executed after the page was delivered.
  - We are deciding which code goes on which page in the server side template and it gets delivered inline with the first request instead of waiting for client side code to execute before retrieving the page specific js

### DIFF
--- a/src/js/rollup.config.all.js
+++ b/src/js/rollup.config.all.js
@@ -26,11 +26,11 @@ const getLastUpdatedFile = (directoryPath) => {
 
 // Combines all the Rollup files into one.
 export default [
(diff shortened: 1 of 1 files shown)


carterm (approved):
Would be great if the rollup file names could be in an array

PR #1697, opened 2020-09-17
Updated CaGov.woff2 font file

kkoryaka:
added 50 more new icons


aaronhans (approved):

aaronhans:
I am not sure why PRs from Konstantin are still getting those Azure build tasks attached. Those are generated from files in .github/workflows which have been deleted and those files aren't present in these branches.

PR #1754, opened 2020-09-22
moved search btn to the right in arabic

kkoryaka:


carterm (changes requested):
Do we still need the html[lang="ar"] if we have the one for dir="rtl"?

carterm:
No longer need with...
https://github.com/cagov/covid19/pull/1766

PR #1766, opened 2020-09-23
adding fix for arabic dir=RTL

carterm:
Making sure css fixes for dir=RTL are included to support Arabic.


carterm:
Pushed to staging

kkoryaka (approved):

PR #1803, opened 2020-09-25
add and suspected to hospitalizations text

aaronhans:


m-sullivan7:
If this uses the existing data point on the live dashboard, I think it's correct as originally written and this change isn't needed.

By contrast, the chart shows confirmed and suspected patients for both hospitalized and ICU patients. This accounts for the discrepancy between the numbers in the dynamic text and the graphs.

Let me know if we should talk this out.

carterm:
Hospitalized confirmed is 2474 while suspected is 833 for this dataset, total is 3307.

This PR makes it incorrect.

aaronhans:
Thanks for clarifying, closing

Issue #1923, opened 2020-10-02
Remove "View all" from the activity dropdown on What's Open search to prevent people from viewing all activities for all counties


adam-little:
Please remove the option to "view all" on the activity dropdown. If a user wants to view all activities for a county, they must select a county and leave this field blank.

adam-little:
![image](https://user-images.githubusercontent.com/69167298/94963168-94119b80-04ac-11eb-972a-075fca732a28.png)

xjensen:
@adam-little This request should now be live on the site via #1970. Just let us know if any changes are needed.

PR #1970, opened 2020-10-06
Remove "View All" from dimmer switch form

xjensen:
Removes the "View all" option from the safer-economy page, per #1923. 

With this change, there's no longer any hint in the UI that people can view all activities. I'm assuming this is a deliberate choice based on the text of #1923 and recent code changes (disabling the card rendering on auto-complete of the county field).

### DIFF
--- a/src/js/roadmap/index.js
+++ b/src/js/roadmap/index.js
@@ -63,7 +63,7 @@ class CAGovReopening extends window.HTMLElement {
     .then(function(data) {
       this.allActivities = data.Table1;
       let aList = []
-      aList.push(this.viewall);
+      // aList.push(this.viewall);
       data.Table1.forEach(item => {
         aList.push(item['0'])
       })


aaronhans (approved):

carterm (approved):

adam-little:
@xjensen this is correct. we want to prevent people from viewing _all activities_ in _all counties_ since there is no use case for this. It also results in a page that displays thousands of search results and breaks the design because we have repeating headers. 

as you point out, there is no longer any hint that people are able to view all activities for their county. 

here are a couple of ideas for how to introduce a hint. what feels achievable to you?

1. Show "View all" in the dropdown _only_ if a user has already selected a county
2. Leave "View all" in the dropdown at all times. If someone selects "view all" without selecting a county, they see an error message that says "Please select a county"

PR #1983, opened 2020-10-08
Removing translations/content folder

carterm:
Removing translation/content folder.  Locations of origin content will be stored in the commit instead.

Wordpress process will no longer add content to the separate translations/content folder.


aaronhans (approved):
nice

Issue #1951, opened 2020-10-05
GET HELP, Health information, Working and How to Help Links are Misleading


kenbasart:
<img width="966" alt="Screen Shot 2020-10-04 at 7 48 46 AM" src="https://user-images.githubusercontent.com/4454420/95106198-0ec8f980-06ed-11eb-9f3b-5eb72bcb7680.png">
<img width="1032" alt="Screen Shot 2020-10-04 at 8 39 38 PM" src="https://user-images.githubusercontent.com/4454420/95106229-1a1c2500-06ed-11eb-8703-bffbbc94c244.png">

Per @clarita1980:

"The titles in the navigation (GET HELP, Health information, Working and How to help) behave like links and 65% of clicks on the navigation are on those titles. It's an interesting fact to know because it's sending the expectation that there might be a page with overall information but because we don't have that it's creating a bad UX. For now, it's recommended to make titles don't behave like links."

cc: @hannahslee

thepegisin:
Suggest those section headers be orange, not underlined, and non-reactive to hover. Only links should look and behave like links.

adam-little:
Here's a screenshot and Figma link for a suggested design. 

Key changes:

- Remove link and underline from section headers
- Make section headers sentence case capitalization (not all caps)
- Page links should not be bold
- Menu button label should read "Close"

![Main menu](https://user-images.githubusercontent.com/69167298/95259939-eed12d00-07f5-11eb-8194-ac6c5705b95f.png)

https://www.figma.com/file/LrzsOu8U5KcMAjJTQ1O3BG/covid19.ca.gov-screens?node-id=2088%3A11

Mark-Tyler:
@kenbasart this is within the skills of Konstantin and not within mine. I would suggest setting this to him.

kenbasart:
Ok will assign it back to him. Thanks!

xjensen:
I saw this and wanted to clarify something, just in case. On mobile, those headers **do need to be links** because they trigger a pull-down menu. This isn't necessary on desktop.

Not expanded on mobile:
![headers](https://user-images.githubusercontent.com/1208960/95400149-c19c8180-08be-11eb-81e3-ab14d34734d8.png)

Expanded on mobile by clicking on the heading:
![headers-expanded](https://user-images.githubusercontent.com/1208960/95400166-c8c38f80-08be-11eb-9fb7-b8f08c246e54.png)

So this raises a couple questions. 

[...]

PR #1985, opened 2020-10-09
Remove expanded-menu clickables from headings on desktop

xjensen:
Per #1951, this removes link-like behavior from headings ("Get Help", "Health Information", etc.) in the header's expanded hamburger menu. 

On mobile, these headings need to be links to allow expansion of drop-down lists of links. On desktop, these links are useless because the lists are already expanded to fill the larger screen. So the challenge is to remove link-like behavior from these headings on desktop, while preserving it on mobile.

Note that these headings still maintain focus-state behavior on desktop with this set of changes.


aaronhans (approved):
nice

PR #2000, opened 2020-10-13
should we manage schools list in WordPress?

aaronhans:


carterm:
Thinking we can manage the tiers in the same file and they can deploy together.

carterm (approved):

PR #2004, opened 2020-10-13
Fixing extra "}" in state-dash

carterm:
Seen between header and title.

https://covid19.ca.gov/state-dashboard/


carterm:
Admin push

PR #2168, opened 2020-10-27
10-27-2020 Weekly Update

carterm:


aaronhans (approved):

aaronhans:
Nice work getting this rolling @carterm!

It looks like we are not going to be able to responsibly eyeball review the updates here since the json is all on one line. I think this is fine because the plan is to move to full automation. It might be cool to get a report to slack when we get to automatic merges, something like the following counties changed the following values...

PR #2175, opened 2020-10-27
fixed map width in blueprint and dash

kkoryaka:


aaronhans (approved):
oh man this is some rough tableau management code, wish these things could scale with percentages

PR #2194, opened 2020-10-29
Adding Pull Request Build Validation

carterm:
This will make sure any PR created for master will have the build run as a check.


aaronhans (commented):
it looks like actions/checkout@master is going to get master even if this is triggered in staging

carterm:
> it looks like actions/checkout@master is going to get master even if this is triggered in staging

Switched to just master for now.

aaronhans (approved):

PR #2280, opened 2020-11-09
roll back contact tracing file

aaronhans:


aaronhans:
A version that was not yet approved by stakeholders got published, rolling back to last approved version

PR #2281, opened 2020-11-09
use global regex flag to avoid error on latest node

aaronhans:
This fixes the build error that happened if you tried to build when running latest node.js version 14

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -344,7 +344,7 @@ module.exports = function(eleventyConfig) {
         
         
         const getNextTag = (searchArea, tag) => 
-           [...searchArea.matchAll(new RegExp('<(?<closeslash>/?)'+tag+'\\b[^>]*>','m'))]
+           [...searchArea.matchAll(new RegExp('<(?<closeslash>/?)'+tag+'\\b[^>]*>','gm'))]
             .map(r=> ({
               index: r.index,
               isCloseTag: r.groups.closeslash.length>0,


chachasikes:
Ran this by checking out `node14` branch, using Node version  `v14.7.0`
Ran `npm install`
Got notification to rebuild node-sass
Ran `npm rebuild node-sass`

Ran `npm run dev` 
The app built successfully.

carterm:
This needs to be tested.  Adding global to a regex makes it return multiple values instead of the first one.

carterm (changes requested):
This works, but it does do a lot of extra work on the regex searching by scanning the entire document rather than just stopping on the first one.  What's the issue here?

carterm:
Looks like the global matchAll is optimized to only bring back what you ask for...so this should be ok after all.
https://javascript.info/regexp-groups

```
Why is a result of matchAll an iterable object, not an array?
Why is the method designed like that? The reason is simple – for the optimization.

The call to matchAll does not perform the search. Instead, it returns an iterable object, without the results initially. The search is performed each time we iterate over it, e.g. in the loop.

So, there will be found as many results as needed, not more.

E.g. there are potentially 100 matches in the text, but in a for..of loop we found 5 of them, then decided it’s enough and make a break. Then the engine won’t spend time finding other 95 matches.
```

carterm (approved):

carterm (approved):

aaronhans:
Thanks for reviewing everybody. Appreciate the testing @chachasikes and code execution research @carterm. 

The reason for this change was that node version 14 refuses to run the build, error message is:

`TemplateWriterWriteError` was thrown
> String.prototype.matchAll called with a non-global RegExp argument

node version 12 has no problem with this but it is nice to fix because version 14 is the active release now

PR #2282, opened 2020-11-10
Translation - restaurants-deliver-home-meals-for-seniors

jorgv:


carterm (changes requested):
KO file has build errors

carterm:
Avantpage is rebuilding

PR #2296, opened 2020-11-10
Remove references to old build:js* package scripts, which are no long…

chachasikes:
Remove references to old build:js* package scripts, which are no longer used because the build tools were recently optimized. CCG-508.


aaronhans (approved):
perfect

PR #2304, opened 2020-11-11
no longer on daylight savings time

aaronhans:
Homepage is displaying wrong update time due to hardcoded DST value, updated to Pacific Standard time offset


abquirarte:
Oops, nice catch

carterm (approved):

PR #2336, opened 2020-11-13
CCG-184 Link County Map button to County Status Map section on Safer …

chachasikes:
CCG-184 Link County Map button to County Status Map section on Safer Economy page. The page id is hard coded, as is the reference to varCountyMap so it makes sense to append the anchor outside of the translated path handler.

### DIFF
--- a/pages/_includes/main.njk
+++ b/pages/_includes/main.njk
@@ -81,7 +81,7 @@
 						</div>
 					</div>		
 					<div class="hero-stats-footer">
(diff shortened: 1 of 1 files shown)


aaronhans (approved):

chachasikes:
Content team reviewed this on staging and decided they want to change the anchor link.

Thanks for reviewing. Good to know this technical approach will work. I may need to change which anchor the link points to & make sure it's available on the instance of the page.

chachasikes:
Content team updated the anchor link, so this page is good to go.

PR #2419, opened 2020-11-20
Ccg 115/editorial control homepage vid

aaronhans:
This is Konstantin's update for the new homepage video under editorial control. I collected this from his old branches and separated it from other stuff that is in different tickets


chachasikes (approved):
Tested these changes in available Desktop browsers, video plays, modal opens and closes, responsive breakpoints between 400-1900 are all good.

PR #2421, opened 2020-11-20
Equity page dynamic data

carterm:
Pulling Top box data from JSON.

Need to update WP file to support this.


aaronhans:
This is already merged to hellofastcharts where it needs to be reviewed and will be merged to master from there with everything else

PR #2422, opened 2020-11-20
translate strings in chart

aaronhans:
This is a PR with an example of retrieving translated strings from the body of the WordPress post. Our constraints are:

- The translation vendor receives content from the WordPress post
- The translation vendor ignores HTML attributes so we can't use data attributes

This requires strings be put in list elements that are children of the custom element with their keys set as classnames

We briefly discussed an alternate approach where we could use a separate table-data post. That would work too and we could end up passing the translation string object into the custom element but the method in this PR seems fine as well.

### DIFF
--- a/pages/wordpress-posts/equity.html
+++ b/pages/wordpress-posts/equity.html
@@ -314,7 +314,19 @@ addtositemap: false
 
 
 
-<cagov-chart-d3-bar></cagov-chart-d3-bar>
+<cagov-chart-d3-bar class="d-none">
+  <ul>
+    <li class="sectionTitle">Factors that increase risk of infection and severe illness</li>
+    <li class="sectionDescription">Californians in crowded housing or transportation, and with less access to paid leave and other worker protections have a higher risk of infection of COVID-19. Social determinants of health, such as food insecurity, lack of health insurance, and housing instability can increase the risk of poor outcomes. These social determinants of health are often the result of structural racism.</li>
+    <li class="chartTitleIncome">Case rate by median annual household income bracket</li>
+    <li class="chartTitleHousing">Case rate by crowding housing</li>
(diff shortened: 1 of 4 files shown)


chachasikes (commented):
Nice!

I was working on the same thing for missingness chart, I can align with your translation class & object naming.

Wondering if using 'class' will potentially conflict with same-name classes in css?

What I decided on (for a case where I need an array of labels.

              
              <ul data-label="displayOrder">
                <li data-label="tests">Tests</li>
                <li data-label="cases">Cases</li>
                <li data-label="deaths">Deaths</li>
              </ul>

Curious to hear your thoughts.

aaronhans:
> data-label

This is a great suggestion. I'll modify my version to use this

jbum:
In the file pages/wordpress-posts/equity.html,   I see a block of markup containing the basic chart text content, which I assume is originally coming from our wordpress, but I don't actually see the corresponding post in our wordpress.  Is it coming from a different word press instance? Is it a hypothetical example that has not been created in wordpress?

jbum:
Also have some questions about this pull request. Is this intended for us to actually pull? (If for us to pull, to pull into what? Our working chart branches?). Or is it just a rhetorical pull request intended to function as departure point for discussion but not actually get pulled?

aaronhans:
> Also have some questions about this pull request. Is this intended for us to actually pull? (If for us to pull, to pull into what? Our working chart branches?). Or is it just a rhetorical pull request intended to function as departure point for discussion but not actually get pulled?

Yeah this was just for a discussion to make sure we were all in agreement on how to get the translated strings for the charts. I'll merge this into the hellofastcharts branch.

The changes to the equity.html were also for illustration purposes. I'll make those updates in the WordPress post

PR #2433, opened 2020-11-23
use env vars to get json location for charts because we have to get d…

aaronhans:
Use environment vars to get the json location for charts because we need to get daily data review before going live so will publish automatically to staging location and not publish to production location until after we get signoff


chachasikes (approved):
Looks good! 
(I need to understand how the rollup-config.js is made available.)

PR #2435, opened 2020-11-24
change Alpine to more restrictive tier

aaronhans:
This is the tier updates PR, opening it now for team review but we should not merge until we are in synch with all other updates


chachasikes (approved):
Looks good!

aaronhans:
After reviewing latest data in snowflake this seems to be the correct update

PR #2448, opened 2020-11-24
CCG-677

m-sullivan7:
This change is to add a link to the Data & tools page, as called for in Adam's Figma design. I used the existing code for the "See what can be open in your county" call-to-action to create this and swapped out the text and the URL as appropriate.

Please stage this change before publishing to production to check that this didn't break anything. Hopefully I placed it in the right spot. Ideally it would be nice to have some space between this link and the "This data drives decisions..." header.

Thank you!

### DIFF
--- a/pages/_includes/dashboard.njk
+++ b/pages/_includes/dashboard.njk
@@ -177,6 +177,14 @@
 
         </div>
 
+        <div class="text-center">
+
+            <a class="button trigger cwdsarrow-parent hi-50" href="/data-and-tools/">
+            <div class="button-label">Explore more data</div>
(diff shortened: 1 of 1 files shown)


aaronhans:
Great work Michael! I am adding Konstantin as reviewer and tagged him on the ticket because he created a cool arrow custom element and I wonder if there is a chance to simplify the markup.

kkoryaka (approved):

aaronhans:
This works great, tested locally to verify nothing else is required to have it work in production branch. Really cool to see new custom elements start being used by editorial. Nice work @m-sullivan7 and @kkoryaka!

aaronhans (approved):

PR #2494, opened 2020-11-30
Corrected case in menu title (COVID-19 and your health).

jbum:


aaronhans:
it looks like Konstantin made this update a few minutes ago https://github.com/cagov/covid19/pull/2493/files

PR #2507, opened 2020-12-01
Task/ccg 637 env yaml update

aaronhans:
Adding a new staging npm script which sets the NODE_ENV
Passing this to the gulp build via command line args in the git action
Use new version of git action which accepts multiple args


chachasikes (approved):
looks good, looks like there is a new build:staging script

PR #2512, opened 2020-12-01
d-none rolls without importance 

aaronhans:
and .chart class is loaded after it with display: block so it wasn't hiding anything


chachasikes (approved):
awesome!!

PR #2526, opened 2020-12-01
Equity page markup fix to prevent double charts and missing charts.

jbum:
There were two errors in equity.html (which I've fixed on WP already). There was a long block of code that was duplicated (the top two charts) due to cut & paste error. There were mismatched tags on the line-chart introduced recently, which I fixed, preventing subsequent charts from displaying.

### DIFF
--- a/pages/wordpress-posts/equity.html
+++ b/pages/wordpress-posts/equity.html
@@ -263,43 +263,43 @@ addtositemap: false
 
 <div class="container">
   <div class="row">
-    <div class="col">
-      <cagov-chart-re-pop class="chart d-none">
-             <ul>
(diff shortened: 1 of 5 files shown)


aaronhans (approved):
Nice, thanks Jim!

chachasikes (approved):
Approving.

One of the commits has Aaron's yaml changes (but it doesn't diff with the main branch)
https://github.com/cagov/covid19/commit/26f3284946fb34225e86199442061cdcfcbf1eb1

Flagging it in case it's weird. Not sure if that's an artifact of merging vs. rebasing, but could be a thing to keep an eye on.

PR #2529, opened 2020-12-01
Deploy/2020 12 1 master to deploy production

chachasikes:
This would be part one of this commit in this manual update process we are trying.


aaronhans:
closing because we merged this here https://github.com/cagov/covid19/pull/2520

PR #2530, opened 2020-12-01
fix for translated pages without translate tag

carterm:
translated pages without the translated tag were getting put in the page root.


carterm:
Works on ...
https://digital-ca-gov.atlassian.net/browse/CCG-736

aaronhans (approved):

PR #2534, opened 2020-12-01
ie11 fixes

aaronhans:
the ie11.css file is imported by the es5.js bundle so it is only used by browsers that execute the type=nomodule script tags instead of the module ones. IE has a problem with svg sizing which is fixed by px hardcoding

The changes to relative-percentage-by-pop/index.js are fixing issues the babel transpiler couldn't handle

The config file is used as an import because it was missing from the es5 bundle and importing from a separate file preserves the code order


chachasikes (approved):
Awesome, looks good

PR #2536, opened 2020-12-02
Fixed county legends on pop chart. Updated WP to match equity changes.

jbum:


chachasikes (approved):
Looks good on initial review.
Good idea to mention the WP update required for the change.

PR #2538, opened 2020-12-02
Feature/2020 12 1 sort ratio

chachasikes:
CCG-719
CCG-735
CCG-640

- Add disproportionality ratio data formula (with notes) - requires a data review & sort by that ratio
- Correct Unknown chart coloring for relative population chart
- Add some conditionals (in as many places as I could get to) to hide data when applied suppression = "Population"

This does not include:
- message box (we need to add a check for all groups = APPLIED_SUPPRESSION === "Total" & then add a message box that fades in. Also need to find an example where this is true, this doesn't apply to Nevada county, need to look through all the smaller counties for examples.
- Grayed out .bar-label (d3 syntax issue)


chachasikes:
Going to rearrange this commit then resubmit

PR #2553, opened 2020-12-02
Changes to display, sorting & incorrect values of relative charts

chachasikes:
- Add disproportionality ratio sort method, with README notes. 
- Correct incorrect chart positions, legends and alignments so that the sort method & data values can be verified and checked. Add applied suppression conditionals and correct incorrect null data for missing data so that we can read the data. 
- Add pseudocode for message box for missing data sets and css classes to use to match Figma design for data with applied suppression.
- Add >= to Statewide/County line
- Add word "County" to chart titles

CCG-719
CCG-735
CCG-640

### DIFF
--- a/src/css/_charts.scss
+++ b/src/css/_charts.scss
@@ -79,6 +79,18 @@
     text-anchor: unset;
   }
 
+  .bars.bars-unknown {
+    font-weight: 100;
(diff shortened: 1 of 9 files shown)


aaronhans (approved):
Excellent! Thanks for making review easier

PR #2556, opened 2020-12-03
Fixed scaling for county-display of dotted line on 100k graph.

jbum:


chachasikes (approved):
Awesome. 
What is the "after we fix it" part? Not sure what's wrong based on the comment.

jbum:
The ideal solution is to provide a separate number for county-specific cases and show that. The current fix shows the statewide number for both county and statewide views (and shows a statewide caption for both). When we get the county number, we'll need to use it, and use the correct caption. The scaling fix is still good though.

PR #2564, opened 2020-12-03
Fix disproportionality sort CCG-719; Fix retina size for 100k chart

chachasikes:
- Remove sort from relative percentage per 100k people chart, sort by METRIC_VALUE_PER_100K instead.
- Fix sort function to simpler calculation of Cases % / Population % for the DEMOGRAPHIC_SET. (Other & Unknown are already excluded from the values for METRIC_TOTAL_PERCENTAGE and POPULATION_PERCENTAGE.
- Move sort function into relative-percentage-by-pop component.

- Fix retina dimensions for relative-percentage-by-pop chart.


aaronhans (approved):
cool

PR #2566, opened 2020-12-03
CCG-770 add county region display to county card

aaronhans:
Ticket: https://digital-ca-gov.atlassian.net/browse/CCG-770

Summary: We have these new regions of the state that include several counties and will be subject to incoming regionwide movement restriction orders. These orders are not yet clear but before noon today we want to associate each county with its region on the /safer-economy page


chachasikes (approved):
Looks good

PR #2567, opened 2020-12-03
CCG-740 Fix extra capitalization

chachasikes:
Move CSS capitalization from .bar-label .tick text to the specific location where we need it (for rendering data that was lower case but needs to be capitalized.


aaronhans (approved):
nice

PR #2574, opened 2020-12-03
added vaccines link, updated tabs mobile

kkoryaka:


aaronhans:
there are a bunch of font file reference changes in here, those are intentional right?

aaronhans (approved):

kkoryaka:
Yes, those are intentional.

PR #2576, opened 2020-12-03
Ccg 225 equity dashboard service for top boxes

carterm:
Puts the real equity data in the top boxes (minus the income)


aaronhans (approved):
nice

PR #2577, opened 2020-12-03
Added aria-label to R/E chart. 

jbum:
Also fixed horizontal alignment of x-axis to match the data line and tooltip.


chachasikes (approved):
looks good

PR #2582, opened 2020-12-03
CCG-677 part 2

m-sullivan7:
Adding spacing beneath the arrow per Konstantin's suggestion. Thanks for the info on Jira. That was super helpful!


jbum (approved):
Looking good!

PR #2583, opened 2020-12-03
Changed titles, added x-axis label on big blue bar chart. 

jbum:
Updated WP to match. Turned off a couple wordy console.logs (turn 'em back on if you need 'em chach).

Note: A styling fix from earlier in the week appears to have broken the highlighting on the tabs for the blue bar chart.


chachasikes (approved):
Looks good.
This change includes code that needs to sync with WordPress.

aaronhans (approved):

PR #2586, opened 2020-12-04
CCG-640 - add missing data message boxes to relative percentage charts.

chachasikes:
Add message boxes to relative percentage charts. 
Needs review with Alpine, Inyo & Nevada counties to double-check our settings & text strings.

@jbum There were some issues with positioning of the text box for the charts, the chart on the left is two svg's stacked. I added a 'faded svg only' to your code to be used in that situation.


aaronhans (approved):
nice code cleanup

jbum (commented):
Looks good!

PR #2594, opened 2020-12-04
Added reusable box (element) tracker. Using it to track all chart views on equity page.

jbum:
This tracker checks if the vertical midpoint of an element has been within the viewport for 3 seconds, then fires a one-time trigger event. The event may be listened to by the tracker to generate a GA tracking event.

```
export default function boxTracker(elementSelector,   // class of element we are tracking 
                           eventName='chart-in-view', // event to trigger
                           labelOverride=null,        // unique name to use, if different than element selector
                           minimumViewMS=3000)        // minimum time in view needed to trigger the event

```


src/js/tracking-you/box-tracker.js
```diff
@@ -0,0 +1,88 @@
...
+                    if (bt_debugging)
+                        console.log(label + " is out of view");
+                    clearTimeout(boxState[tKey].lh);
+                    boxState[tKey].inView = false;
+                }
+            }
+        }
+
+    };
+    window.addEventListener('scroll',throttle(scrollHandler,1000));
```
aaronhans, line 87: cool, glad to see throttling implemented with a relaxed timer

aaronhans (approved):
cool feature, nice code organization

chachasikes (approved):
Nice work!

PR #2595, opened 2020-12-04
Center relative population chart message box (CCG-707)

chachasikes:


jbum (approved):
A masterpiece of CSS. Chef's kiss.

PR #2597, opened 2020-12-04
Update equity page content

chachasikes:


chachasikes:
Sync Wordpress changes - Please review that this had all your changes @jbum 
@aaronhans may also be editing this same file

jbum (approved):
Looks different but harmless.

jbum:
The learn more section has been replaced with some explicit markup that I think was being generated in js (I forget why). Will clean up the js side if needed after this merge.

chachasikes:
I think Aaron got this already with other updates, so gonna close it to be safe. @aaronhans

PR #2600, opened 2020-12-04
CCG-338 Remaining p0 issues for equity analytics tracking.

jbum:


jbum:
Re-ordered params on box-tracking for convenience.

src/js/tracking-you/index.js
```diff
@@ -296,6 +296,22 @@ export default function setupAnalytics() {
         reportGA('county-select-typo', e.detail.county, 'activity-status');
       }.bind(this), false); */
       
+      // Setting up trackers for big blue bar chart
+      let large_tab_classes = ['income','housing','healthcare'];
+      setTimeout(function() {
```
aaronhans, line 301: this setTimeout doesn't seem like the optimal solution

Can we try something that will be more reliable? Some options could be: adding listener on body and looking for parents with specific class or checking for elements to show up in a loop that tries a few times

jbum: OK :)

PR #2603, opened 2020-12-05
Analytics fixes. Including a better way of delaying tab tracking.

jbum:
Using a custom event at the end of the chart constructor that is creating the tabs that are being tracked.


aaronhans (approved):
right on

PR #2607, opened 2020-12-05
Rsho

aaronhans:
Regional Stay Home Order

Get new list of regions in lockdown
Use this on the What's Open search
If a county is in a region which is in lockdown (RSHO)
then display the activity restrictions from the new RSHO column in the reopening activity matrix dataset instead of the county tier activity info
This PR also gets some defensive coding that was in the roadmap/index.js file in staging


aaronhans:
This should be left open until we have stakeholder approval to release possibly Sunday 12/6

chachasikes (approved):
Looks good & tested locally since there were a bunch of changes to master

carterm (approved):

PR #2608, opened 2020-12-05
change how we are calculating stuff

aaronhans:
I think this fixes the data in the top boxes but this should be reviewed

Fetch the data for the top box info in the js file that provides it to 11ty
Remove the request from gulpfile
We now have a clean statewide rate provided in the income_cumulative query, use that everywhere
I will uncomment the same thing in WordPress so ignore that change


chachasikes (approved):
looks good

carterm (approved):
Works, as long as we are ok doing the fetch with each build, which we are doing in other places.

PR #2609, opened 2020-12-06
Code cleanup. Fixed bottom margins on big blue chart for mobile. Minor tweaks for Spanish on line chart.

jbum:


src/js/tracking-you/index.js
```diff
@@ -298,18 +298,10 @@ export default function setupAnalytics() {
...
-            console.log("Found",selector);
-            btn.addEventListener('click', function (e) {
-              reportGA('tab-select', tlabel);
-
-            });
-          });
-        });
+        for (let tlabel of ['income','housing','healthcare']) {
+          const btn = document.querySelector("button.large-tab." + tlabel);
+          btn.addEventListener('click', (e) => reportGA('tab-select', tlabel) );
```
aaronhans, line 303: seems like this is vulnerable to an error if the element is not found when the other version wasn't because the selector is defined inside the loop and btn could be undefined when you try to access .addEventListener. Could add a check to make sure it exists first

jbum: Sounds good.

PR #2613, opened 2020-12-07
Temporary fix for RTL languages (Arabic) for equity page. CCG-815

jbum:


chachasikes (approved):
Checked this out & looked at the -ar page.
Look like the RTL placement for the chart label is aligned better.

aaronhans (approved):

PR #2625, opened 2020-12-08
CCG-816: Repaired korean translations. Localized date display on line chart.

jbum:
Korean translations will also need repair on master. 
Several instances of _placeholderForDynamicLocation_ were changed to _placeholder_.


chachasikes (approved):
looks good

PR #2636, opened 2020-12-08
CCG-647 RTL Override for search div. Generalized rtlOverride utility.

jbum:


jbum:
This also contains a fix the date-localization code.

chachasikes (approved):
looks good

PR #2640, opened 2020-12-08
changed default size number so map is not too big on small screen

aaronhans:
This swaps in a new tiers/RSHO map version and makes some updates to the chart widths to accomodate that new tableau map


aaronhans:
this update is in staging

chachasikes (approved):
thumbs up!

PR #2641, opened 2020-12-08
CCG-810. Improved tooltip language for r/e pop chart. Displaying the same tooltip for both charts.

jbum:
I also changed the corresponding language on WP.


chachasikes (approved):
nice

PR #2643, opened 2020-12-08
Syncing with Blueprint

m-sullivan7:
Adding language around the county map that's present on the Blueprint page for the sake of consistency.


jbum (approved):
Looks good!

PR #2645, opened 2020-12-09
Task/ccg 829 e2e test

aaronhans:
End to end test that reviews the blueprint page and makes sure the What's Open search functionality works

This is integrated into the existing build validation workflow that runs on opened PRs so it won't slow down content updates but will run against code changes


chachasikes (approved):
Awesome!!!

jbum (commented):
This looks very cool. I'll need to spend a lot more time with it before I'm knowledgeable enough to approve it (my 11ty and puppeteer experience is still nil). But I approve in principle!

PR #2657, opened 2020-12-09
fix math

aaronhans:


chachasikes (approved):
Important correction.
Maybe we can start to separate out our formulas like with the disproportionality ratio - having a clear formula from data scientists, a test data where we confirm the formula & a clear story & data definitions would help - especially if different developers come in & need to catch up on what the data means.

aaronhans:
Great point, thank you @chachasikes I created https://digital-ca-gov.atlassian.net/browse/CCG-862 to make that upgrade

PR #2668, opened 2020-12-10
add a unit test

aaronhans:
This is a unit test of an 11ty data processing function that ensures the data returned is as expected given by overriding the fetch function via proxyquire, feeding in a canned set of data and comparing to the expected return values


jbum (approved):
Looks like a useful example. Thanks!

chachasikes (approved):
Awesome. I like the mocks folder.
Requires npm install & then npm run testdata to set the test suite in action.

PR #2674, opened 2020-12-10
updated workflow with our standard SWA args

aaronhans:
Added missing workflow task arguments that should make this site browseable


jbum (approved):
Looks good.

PR #2683, opened 2020-12-11
Added missing captions to translated Equity page files that were missed last translation pass.

jbum:
This fixes a problem in which an informational message about low-population counties doesn't show up on the right side for translated pages.


aaronhans (approved):
thanks Jim! Appreciate all your work making sure the data is displayed correctly through all translated versions.

Thanks for being proactive with this fix and making sure it remains when we get the next batrch of translations from the vendor

PR #2690, opened 2020-12-11
CCG-59 Fixed text alignment issue in Firefox. 

jbum:
Note: Firefox has a more restrictive use than Chrome of which tags can take alignment-baseline, but can use dominant-baseline for the same purpose.

Added vestigal testing directory to .gitignore.


chachasikes (approved):
Looks good

aaronhans (approved):
nice

PR #2693, opened 2020-12-12
CCG-868. Workaround for date on equity. Instead of WP date, use yesterday's date.

jbum:
Temp fix while we come up with something better. Given daily updates, the typical date is going to be yesterday, since the data is processed overnight on yesterday's data.


aaronhans (approved):
ship it

PR #2696, opened 2020-12-12
get county select working again

aaronhans:


chachasikes (approved):
This is an improvement, some scoping issues & issues of the chart worksheet being available when the county request happens.

There is a delay while the tableau content is loading, but once the charts stop processing, the toggles work. I recommend that we merge this change in to make the toggle button work at all (even if it is slow.)

jbum (approved):
Agreed :)

abquirarte:
☕👀 interesting!

PR #2710, opened 2020-12-15
Comment out test that is blocking successful builds. 

chachasikes:
We need to look into what's wrong, but it's blocking development & the code does not appear to be actually broken.

### DIFF
--- a/src/js/roadmap/countystatus.json
+++ b/src/js/roadmap/countystatus.json
@@ -53,7 +53,7 @@
   },
   {
     "county": "Inyo",
-    "Overall Status": "4"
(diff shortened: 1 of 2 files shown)


jbum (approved):
Ship it

carterm:
This will deploy the tier change! we ready for that?

src/js/roadmap/countystatus.json
```diff
@@ -53,7 +53,7 @@
   },
   {
     "county": "Inyo",
-    "Overall Status": "3"
+    "Overall Status": "4"
```
carterm, line 56: Ready for this?

PR #2712, opened 2020-12-15
Add whitespace for pinging server process

chachasikes:


jbum (approved):
Excellent use of whitespace.

PR #2717, opened 2020-12-15
commit stamp test (Do not deploy)

carterm:
(Do not deploy)

Experimenting with builds recording their commit so we can look for them during deployment.


carterm:
resolved with...
https://github.com/cagov/covid19/pull/2811

carterm:
Ok it is in the build now.  Don't need this one anymore

PR #2723, opened 2020-12-16
Error checking to prevent error in which clicking to the left/right of tabs triggers an invalid update.

jbum:
The entire div that holds the buttons on the equity page (cases/tests/deaths) generates click events, but only the buttons have valid data. Clicking to the left/right of the buttons triggered an invalid chart display. 

Checking that there is valid data before triggering a page re-display.


jbum:
CCG-985

coreycaitlin:
🎉 I checked, and it looks like this fixes it on the lemon-pond preview site. I'll defer to someone else for code review, though.

abquirarte:
🍋💃

chachasikes (approved):
Looks good.

PR #2726, opened 2020-12-16
adding commit message to build

carterm:
This should change the commit message in the Azure build to match the commit message.  Should help with troubleshooting.
(Also removing an old ignore reference)


chachasikes (approved):
This is an awesome addition.

PR #2731, opened 2020-12-16
Inhibiting initial display of county map on state dashboard.

jbum:
Charts on state dashboard are currently displaying one on top of another (Statewide on top of County) since a tableau timeout is causing an error which shadows some code. I'm inhibiting initial display of the county map to prevent this visual glitch.


chachasikes (approved):
Sounds good - Jim says that this change hides something that was hidden anyways, so we shouldn't have to worry about chart being hidden when the tableau service is back up

PR #2734, opened 2020-12-17
CCG-1002 Line chart x-axis fixes. Fixed date parsing.

jbum:
Added utility parseSnowflakeDate to be used instead of new Date('2020-12-05') which prevents the date from being interpreted as GMT, which produces off-by-one-day issues. Increased right margin and adjusted offsets on X-Axis.


chachasikes (approved):
Nice

PR #2742, opened 2020-12-17
Bug/2020 12 17 suppression

chachasikes:
This resolves an incorrect labelling issues for the Relative Percentage by Population chart.
* adds new labels for bars for situations where 'total' and 'population' data withholding/applied suppression are applied (situations with fewer than 11 deaths or populations with less than 20,000 people)
* adds css class to set the size of the label.

Note: This does not have text area sizing & we may need to address this with design for translated content or reduce the length of the text.

equity.html page has been updated already to connect to this code & strings for different cases, deaths and tests set up.


coreycaitlin:
@chachasikes Looking at Calaveras, as an example, it looks like we're missing the <race/ethnicity> before "people". So "Data is not shown because there are fewer than 20,000 people in this county." should be "Data is not shown because there are fewer than 20,000 <race/ethnicity> people in this county."

This will definitely trigger multi-line issues, as you noted in Slack.

jbum:
The chart on the right side (100k) may still need tooltip adjustment.  100k related fields are suppressed.  But I'm seeing this tooltip

image

PR #2760, opened 2020-12-18
Adding rel="external" to news links

carterm:
The news feed is full of links to other sites.  These need to be annotated as external to prevent SEO confusion, and let robots know that those links aren't ours to maintain.

Attempts to resolve...
https://digital-ca-gov.atlassian.net/browse/CCG-1031


chachasikes (approved):
great!

PR #2777, opened 2020-12-21
Switched Latino box from cases to deaths 

jbum:
CCG-1026

Matching methodology used for African-americans in third box. 

Will update Wordpress to match when this gets approved.


chachasikes (approved):
Looks like you have added this field data:

death_rate_vs_statewide_percent_latino
death_rate_per_100K_latino

and set the values in the data 
and then updated the equity.html page with  death_rate_vs_statewide_percent_latino & set deaths values.

PR #2783, opened 2020-12-22
CCG-776 Fixed tooltips in firefox. Removed some console logging.

jbum:
Note: event.offsetY/X works differently in Firefox than Chrome.  

event.layerX/Y is a better choice and more closely corresponds to mouse position for element repositioning.


chachasikes (approved):
looks good

PR #2800, opened 2020-12-22
Expand x-axis by one-day on each side. Looks nicer. No clipping.

jbum:


jbum:
image

chachasikes (approved):
Nice, the right edge is less smushed now.

PR #2811, opened 2020-12-23
Adding a "commit-info" message

carterm:
This should create a json file with the commits included in the deployment.  We can use this to verify deployments.


chachasikes (approved):
Curious: where will this message appear? @carterm

carterm:
Should just create a file in the root.

carterm:
Holding off to deploy this for a bit until i can watch it.

PR #2815, opened 2020-12-23
CCG-808 Enlarged hover target areas for Pop, 100k and Missingness (skinny bar charts).

jbum:
After drawing a bar, we draw a second bar on top of it which is transparent, but 20 pixels thicker, and is used for both the aria label and the hover-area for tooltips.  

In the case of the pop chart (which has two bars per item), we use a single large hover area over both of the skinny bars.


chachasikes (approved):
Ooh! Variables! :)

PR #2816, opened 2020-12-23
proper file path for commit info

carterm:
didn't work last time.  adding the "docs" prefix


chachasikes (approved):
yay for correct paths :)

PR #2837, opened 2020-12-28
CCG-908 Tweaked positioning and size of x-axis labels and caption.

jbum:
This corrects some visual glitches for the bar chart on mobile and tablet.
Increased bottom margin to account for changes in aspect ratio, so X-Axis label doesn't collide with tick text.
Increased size of SVG for tablet and non-retina mobile.
Changed aspect ratio of retina mobile to match better.
Adjusted tick text sizes for mobile and tablet for readability and proportions.


aaronhans (approved):

chachasikes (approved):
awesome

PR #2847, opened 2020-12-29
CCG-908 Additional chart tweaks for mobile.

jbum:


jbum:
Added a hack which checks the relative proximity of the 3rd and 4th bars in the chart. If the 4th bar is closer, we left-align the "Statewide cases" caption. Otherwise, we right align it as usual.

chachasikes (approved):
Lots of tweaks! :)

PR #2851, opened 2020-12-29
Task/ccg 481 kk feedback footer

aaronhans:
This is the update to the page feedback component in the footer created by Konstantin

We have abandoned the published component because that was a universal widget that could have been used on other sites. Now we have something that reads in ethnio urls and is only appropriate for use on the covid site.

- Translated strings come from WordPress, in the common-page-labels post which has been published to staging
- Those strings are used in the njk template and fed into the component as data attributes

- The component still collects info typed into a form field on the page that is shown if you say you did not find what you need
- This info goes to the Azure CosmosDB as it did in the last version of the component
- Links to different ethnio surveys are presented based on whether you said you found what you wanted or not

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1644,11 +1644,6 @@
         "awesomplete-es6": "^1.2.2"
       }
     },
-    "@cagov/pagerating": {
-      "version": "2.0.0",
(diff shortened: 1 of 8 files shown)


chachasikes (approved):
I see that we removed a page rating npm package, added a new web component (which looks nice by the way)

Awesome pull request with ticket number & a summary of the change!

PR #2854, opened 2020-12-29
CCG-1112 Stacking legends. Other mobile fixes.

jbum:
Stacking legends on pop chart vertically when width of chart is < 400 to avoid text cropping on mobile.
Positioning fix to avoid legend cropping on line chart.
Better proportions on line chart on retina.


chachasikes (approved):
Approved!

aaronhans (approved):

aaronhans:
nice

PR #2858, opened 2020-12-30
add some documentation for css purging

aaronhans:
CSS purging confusion comes up sometimes, wrote some docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ We need to balance pregenerating a lean site with ease of authoring. We do this 
 
 # Development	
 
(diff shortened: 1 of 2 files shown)


chachasikes (approved):
Docs!

jbum (approved):
Thanks! Curious what other differences there are between npm run dev and npm run start. Guess I'll take a look...

aaronhans:
> Thanks! Curious what other differences there are between npm run dev and npm run start. Guess I'll take a look...

- The javascript transpilation to support IE11 doesn't get run in dev mode
- The js and css in each page are referenced in script and link files in dev mode instead of inlined into the HTML. This lets js and css changes reload right away instead of requiring a new 11ty build.

PR #2865, opened 2020-12-30
CCG-1094 Added suppression captions to 100k chart for consistency with pop chart.

jbum:


jbum:
Still evaluating multi-line caption options.

chachasikes (approved):

PR #2866, opened 2020-12-30
CCG-208 Updated English title of latest-news page.

jbum:
Quick English-only patch. There is some internal discussion of how to deal with translations on manual-generated pages.


coreycaitlin (approved):
👍

PR #2871, opened 2021-01-01
Fixing 2020 GMT issue

carterm:
So I guess we never thought this would go into next year...


aaronhans (approved):
thanks for sorting this out

PR #2875, opened 2021-01-01
trying to fix dash date issue

carterm:
FIXED support for dates without GMT codes

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -232,9 +232,7 @@ module.exports = function(eleventyConfig) {
         ? datestring 
         : datestring==='today'
             ? new Date()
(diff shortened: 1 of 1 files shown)


carterm:
trying in staging first...
https://staging.covid19.ca.gov/state-dashboard/

chachasikes (approved):
Carter removed some extra data formatting code that's complicated & seems to make the year incorrect. It does render correctly. 
May also have something to do with 'addDays' & error may also have appeared on guidancefeed related data.

chachasikes (approved):
Extra fix to accommodate timezone - confirmed dates correct on state dashboard & safer-economy

PR #2879, opened 2021-01-03
1-3-2021 Stats Update (staging)

wpserviceuser:


Monesha67:
The daily cases on the COVID-19 California incorrect on the website because doesn’t match with JHU-CSSE, Worldometer, CDC website. Please update the daily cases and the deaths count. Here’s the link to Worldometer. https://www.worldometers.info/coronavirus/country/us/    Thank You!

PR #2891, opened 2021-01-05
remove inline style

aaronhans:
This looks like an innocuous cleanup change but there is something more sinister going on. 

- We are passing height and width values to tableau when we ask it to display a chart but this is not enough to fully control layout. If a chart is loaded but not displayed tableau will get confused about layout modes to choose causing fallback to mobile view. 
- We are also seeing a bug in Firefox where loading a chart when not displayed causes tableau to fail to set the active worksheet on the active workbook?!? This will break county search. 

I conferred with Jim to make sure this code change won't wreck some other important bug fix. It was put in place to help with the flashing of unwanted county chart which can happen on load. Since we have had that bug since launch but the layout and FF problems have been raised as high priority we will remove this display:none

### DIFF
--- a/pages/_includes/dashboard.njk
+++ b/pages/_includes/dashboard.njk
@@ -175,7 +175,7 @@
         <div class="row d-flex justify-content-md-center">
 
             <div class="toggle-group-container cases-group">
-            <div  id="cases-county-graph" class="toggle-group-element" style="display:none;">
+            <div  id="cases-county-graph" class="toggle-group-element">
                 <div class='tableauPlaceholder' id="casesChartCounty"></div>
             </div>
             <div id="cases-state-graph" class="toggle-group-element">


chachasikes (commented):
"sinister" (!)

Ran this locally and county search was still working... but reading the Jira ticket - I can see that there are some cross browser tests. County search does work on preprod in Firefox.

chachasikes (commented):
Need to QA "IE11 on Win10 via browserstack: Totally broken, no charts display"

chachasikes (approved):
This doesn't effect IE11 ... but have a feeling that's a bigger fish to fry - so approving this so that the Firefox county search will be fixed. 
Tested on Chrome, Firefox, Safari iPhone & charts all look good.

aaronhans:
Yeah, we keep having regressions on IE11. It is so easy to make a change that breaks stuff in that browser and we aren't doing full cross browser reviews on every update. I opened this ticket to fix IE11: https://digital-ca-gov.atlassian.net/browse/CCG-1199

aaronhans:
Thanks for reviewing @chachasikes!

PR #2904, opened 2021-01-05
package updates for security and dependabot bumps

aaronhans:
Package bumps 

These changes have already been tested in staging

- Merging all recent package bumps from dependabotprs branch
- Moving a couple build dependencies into devDependencies (thanks @carterm)
- Using force resolutions to bump a deep dependency to fix high security vulnerability (npm install --depth didn't work)


chachasikes (approved):
Approved.
Install worked with regular dev & build processes.

PR #2906, opened 2021-01-05
remove inline style

aaronhans:
PR to preprod: https://github.com/cagov/covid19/pull/2891

The inline style on this element makes tableu layout the county cases chart incorrectly and also causes tableau to fail to populate the active worksheet (only in Firefox) which breaks the county switch in that browser.


chachasikes (approved):
Looking good

PR #2905, opened 2021-01-05
updated accordions look and feel

kkoryaka:

PR #2907, opened 2021-01-05
updated accordions look and feel

aaronhans:
doing this one in preproduction first https://github.com/cagov/covid19/pull/2905


chachasikes (approved):
RE: CCG-825
Approved to push to preprod for QA (but have some issues running this locally) - we are looking into it so let's not merge to master until we are sure - otherwise we might be breaking all the site accordions. (Could be a merge conflict.)

PR #2921, opened 2021-01-06
remove now unused sha creation and storage

aaronhans:
This change needs to be reviewed with the setting: "Hide whitespace changes"

- The creation of the hash and the code to store hashes for comparison has been removed

We removed the if statement that did the hash comparison in Konstantin's PR that updated accordions when we discovered a bug with accordion processing and that this little hash scheme wasn't helping build times at all. Removing the remaining now unnecessary code prevents a little unnecessary processing


chachasikes (approved):
great - I see you removed the other related code that's no longer needed related to our accordion changes yesterday.

PR #2922, opened 2021-01-06
Task/ccg 1195 data link callout

aaronhans:
Adding more dramatic styling to link to data and tools page

As part of or right after the state dashboard 2 launch we need to address all this content being here instead of under usual editorial control in WordPress


m-sullivan7:
Good on my end! I probably shouldn't be the person to merge the pull request--but can be if needed!

chachasikes (approved):
Looks nice

PR #2927, opened 2021-01-07
Task/ccg 1195 update feedback flow

aaronhans:
Updating the feedback flow to request comments in the existing CosmosDB powered form for positive initial button presses as well as negative ones. This seems like a nice change.


chachasikes (approved):
Looks good!

PR #2928, opened 2021-01-07
remove now unused sha creation and storage

aaronhans:
Moving the 11ty build optimization to production branch. 

To review make sure hide whitespace setting is on

The change removes the sha creation and storage logic that is no longer used after we discovered it doesn't speed things up so removed the check against the sha in the last accordion update which is already live.


chachasikes (approved):
Removing some extra code connected to a removed preprocessing script.

PR #2948, opened 2021-01-08
CCG-908: Reduced caption size on intermediate sizes (520-768). 

jbum:
Also reduced overall font-weight at smaller sizes, and using n-dashes instead of space-hyphen-space to reduce caption size.

This fixes caption overlap in the 500-600 width range on Firefox and Safari.


chachasikes (approved):
Don't see a ticket number, so I assume this was trying to resolve a prior congested text issue on mobile devices? Normally I would sound an alarm about small size text. The text is starting to get small but it's still zoomable on iphone. Usability wise, we may actually want to, in the future, consider mobile chart variants - that are not inheriting too much, but actually full on different layouts, or push to design for mobile first next time. But for now, if this addresses the issue I think it does, I am ok with this change. :)

PR #2973, opened 2021-01-13
Bug/ccg 262 progressive enhance please

aaronhans:
Problem: https://digital-ca-gov.atlassian.net/browse/CCG-262 there is a bug in legacy Edge versions 17 & 18 where they download all the javascript, both module and nomodule versions and then fail with errors that break all web components.

There doesn't seem like a good solution for that legacy browser (Which is scheduled for EOL in March 2021 per Microsoft.

We can make the accordion component fail gracefully so that it doesn't prevent people from reading stuff though. The following changes were made to accomplish this.
- The accordion component was republished with a major version bump with a line added to add a new class to the custom element
- The accordion version used by the site was updated
- The custom element was changed to use cagov prefix everywhere
- CSS was changed to let the component content be visible if all javascript fails and for the accordion functionality to enable as js executes. The base js was already inlined on this site so the browser knows to close the accordions right away
- The polyfill we used to load to allow legacy Edge to support web components was removed. This browser's script tag behavior is so buggy we are only going to be able to provide very basic support. I don't recommend putting a lot of time into this specific browser due to its impending EOL


jbum (approved):
:+1:

PR #2983, opened 2021-01-14
kk-survey-st content

aaronhans:
This is Konstantin's implementation of the new survey. Clara has reviewed this in staging.


chachasikes (approved):
Adds a survey about vaccines & merging to preproduction for more review.

I see two surveys on the main include file.

Ran the code but didn't trigger viewing the survey. Would be good to do an extra QA on preproduction.

aaronhans:
Thanks @chachasikes will do another round of QA in preproduction. It looks to me like it is working as desired, the survey is presented on a small percentage of page loads and once you've seen it a flag is set in localstorage and it is not presented again

PR #2989, opened 2021-01-14
Task/ccg 1281 schools may open text

aaronhans:
New PR to move the text for school reopening from the js into WordPress


aaronhans:
have to go live now

PR #3014, opened 2021-01-19
CCG-1332

m-sullivan7:
Update to state dashboard to include the vaccines dashboard. Please stage prior to moving to production. :)

@jbum once we have the final production code (it's entirely possible this won't look like we want on the first try), we should add this code to the state dashboard 2.0 version as well. I'll let you know when we're ready for that!


carterm:
Pushed to staging

m-sullivan7:
@carterm Good to merge and push to production!

carterm (approved):

carterm:
...and away it goes...

PR #3034, opened 2021-01-21
CCG-1273 CCG-1348 Added vaccine data. Changed daily tests data source.

jbum:


chachasikes (approved):
Adding new fields to interface

PR #3049, opened 2021-01-21
Ccg 1287/bug event tracking prefixes

aaronhans:
Problem: Britt noticed that the homepage want to know click events stopped tracking clicks.

Cause: The event handler was no longer looking at a class that in use

Solution: We can fix this by following the convention of using class hooks with the js- prefix for event hooks. We shouldn't attach style rules to these classes, only use js-whatever for javascript. Then it will be easier to notice if event hooks are being changed when devs are modifying HTML.

Changes:
- Switched class references to use js- prefixed classnames for the events which were very specific in the homepage event tracking'
- changed the classes to match on the page and in component
- Removed commented out code

I documented the js- prefix convention in the event tracking wiki: https://github.com/cagov/covid19/wiki/Tracking-Events

### DIFF
--- a/pages/_includes/main.njk
+++ b/pages/_includes/main.njk
@@ -242,7 +242,7 @@
 						<ul class="arrow-links">
 						{%- for item in items %}
 						<li>    
-							<a class="link-arrow-white" href="{{ item._url }}">
+							<a class="link-arrow-white js-event-hm-wtk" href="{{ item._url }}">
 								<div class="link-arrow-label">{{ item.text | safe }}</div>
 								<cagov-arrow></cagov-arrow>
 							</a>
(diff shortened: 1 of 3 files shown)


aaronhans:
Build fail due to Azure dynamic env limits: "Reason: This static site already has the maximum number of staging environments (3). Please remove one and try again."

chachasikes (commented):
I like this naming convention idea & it makes sense.

Question: what is "js-event-hm-wtk" - not sure what hm is or wtk - are they connected to another service?

carterm (approved):
Looks like you added some css and removed a commented section.

aaronhans:
> I like this naming convention idea & it makes sense.
> 
> Question: what is "js-event-hm-wtk" - not sure what hm is or wtk - are they connected to another service?

Maybe I should make the class name more expressive. The classname links the home page want to know links to the google analytic event report function call that grabs the href and uses it as the event label.

PR #3051, opened 2021-01-22
Ccg 1362/bug in messaging

aaronhans:
The prompt after a negative response was chosen was inappropriate because the same data attribute name was used 

Fixed the attribute name fixing the bug

ticket #CCG-1362


jbum (approved):
Looks good!

PR #3052, opened 2021-01-22
fix the strings passed into feedback widget

aaronhans:
This has been merged to preproduction and tested there, wrong string was being presented during feedback flow


chachasikes (approved):
Changing a prop

PR #3075, opened 2021-01-25
CCG-1360 Number formatting fixes on equity page.

jbum:
Corrects the 100k and big blue bar chart to use integers with comma delimiters for sample numbers.


jbum:
Note: This was done in a such a way so that we have the option of using localized number formatting in the future, although I suspect it's not actually gonna be needed here.

chachasikes (changes requested):
Nice call on the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

My build of this breaks something...

"built.js:1 Uncaught TypeError: Cannot read property 'nextElementSibling' of null
    at built.js:1"
    
    Not sure what's causing it
    
    Noticed that the language picker wouldn't load.
    
    Javascript isn't totally broken on the page, but wondering if there is an issue

jbum:
Hey Chach,

That error appears to have already been in preproduction (and staging). It's triggered by this line, which I think is related to the dark-accordion work that @kkoryaka is working on.

p.nextElementSibling.classList.add("dark-accordion-sibling"),

PR #3080, opened 2021-01-26
checking if dark accordion element is available on the page to avoid …

kkoryaka:
…undefined error


jbum (approved):
Looks much better. Thank you!

PR #3097, opened 2021-01-27
fixed most of dashboard page SiteImprove a11y errors

kkoryaka:


aaronhans (approved):
Thanks for fixing the a11y issues!

chachasikes (approved):
Looks good

PR #3111, opened 2021-01-27
fixed most of blueprint page SiteImprove a11y errors

kkoryaka:
Awesomeplete component's errors are still in there (awesomeplete needs separate package update)
Tableau iframe error is still there. Not able to fix that at the moment since we don't have control of Tableau

### DIFF
--- a/src/js/roadmap/template.js
+++ b/src/js/roadmap/template.js
@@ -10,6 +10,7 @@ export default function template(json) {
             <input
               aria-expanded="false"
               aria-owns="awesomplete_list_1"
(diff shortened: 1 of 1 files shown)


jbum (approved):
Thanks for addressing the SiteImprove stats!

carterm (approved):
looks like you added some aria attributes, change a label name,  and removed 2 hidden ULs.

aaronhans (approved):
thanks for identifying the cause and figuring out these elements were duplicates of what awesomplete creates dynamically

PR #3119, opened 2021-01-28
remove the dashboard code from the IE11 bundle

aaronhans:
Yep, just commenting out the dashboard code in the IE bundle

There are several errors in this code that breaks that page specifically, those will be addressed in CCG-1411. The CCG-1410 ticket prevents that problematic code from also breaking other pages in this browser


aaronhans:
need to merge to test and deploy this production fix

PR #3122, opened 2021-01-29
remove the dashboard code from the IE11 bundle

aaronhans:
This page is broken in IE11, this code change fixes it and unfortunately does not break state dashboard anymore than it already is in IE11.

This change has been tested in preproduction


chachasikes (approved):
thumbs up

PR #3127, opened 2021-01-30
fix tracking events

aaronhans:
events got left behind again, this code change fixes the reporting, will take longer to write proper tests so will address that in CCG-1424


jbum (approved):
Looks good!

PR #3129, opened 2021-01-30
Vaccine boxes changed on homepage and /state-dashboard/

jbum:
Includes /v2-state-dashboard/ code which is currently not active due to redirect page on wordpress.


aaronhans (approved):
cool

chachasikes (approved):
still looks good

PR #3140, opened 2021-02-01
fixed ID is not unique a11y error on the home page

kkoryaka:


aaronhans (approved):
Do we need these ids? Are they useful? We should be assigning any event listeners to js- prefixed classes and any styles to standard classes. 

If the id values can be changed with no consequence maybe they can be removed.

kkoryaka:
You are right. It looks like we don't need those IDs

PR #3143, opened 2021-02-02
added missing blue color class to the general.scss

kkoryaka:
so, editors can easily add .color-blue class to the text blocks in wp posts


aaronhans (approved):
blue classes, cool

PR #3187, opened 2021-02-05
Jbum minimized

jbum:
Using minimized version of tableau-2.js, which greatly reduces noise in the console and loads faster.  Thanks to Carter for noticing this.

Reverted arrows on homepage and dash to their former state (forced upwards).


chachasikes (approved):
Nice - glad there's a minified version of tableau.

PR #3196, opened 2021-02-05
fixing timeout for puppeteer

carterm:
This should fix a lot of the automated testing errors.  Some processes are taking 32+ seconds to run, this will set the timeout to 60s along with the other timeouts.


aaronhans (approved):
thank you Carter

PR #3215, opened 2021-02-09
added important to the icon display property when accordion is open

kkoryaka:


kkoryaka:
Needed to add !important to the icon display property when accordion is in open state. Otherwise pluses and minuses don't toggle.

aaronhans (approved):

PR #3226, opened 2021-02-09
replaced CaGov plus/minus icons in accordions with SVG code (ccg-1470)

kkoryaka:
Moving new SVG accordion plus/minus icons replacement to production


chachasikes (approved):
Nice!

PR #3227, opened 2021-02-09
Ccg 1497/update survey frequency

aaronhans:
Show the vaccine survey less frequently at Clara's request


chachasikes (approved):
This makes the survey appear less frequently?

PR #3233, opened 2021-02-10
Ccg 1437/fix data collection event

aaronhans:
- Fix test that was reporting homepage menu clicks
- Add tests for that GA event and the last broken one: feedback clicks


chachasikes (approved):
Non blocking notes:
* requires `npm install`
* could `waitForevents`  be `waitForEvents` (consistent camel case)?
* what's the `$$eval` ?

aaronhans:
Good points
We probably want to warn everybody when we make an update that will require an npm install on their next local branch.
camelCase fixed
$$eval returns a promise that resolves to an array of the selector, I don't need the return value here though so I swapped them out for waitForSelector whose purpose seems more obvious

PR #3240, opened 2021-02-10
Centering tableau chart per Karim's request.

jbum:
Note for Chach: This is an edit to our modified copy of the vaccines wordpress file.  I already fixed this in Wordpress.


aaronhans (approved):
cool

PR #3241, opened 2021-02-10
2021 02 10/vaccine text

chachasikes:
* Add uncentered arrow (ready for styling)
* Update & correct display text to match recent content choices
* Update data labels & input/outputs
* Add "County" after county name in chartTitle
* Update text strings to use translatedObj strings

NOTE: using [REGION] as placeholder text, though the preference is this:
<span data-attribute="county">[REGION]</span>

This syntax is a little tricker to check for, because you need to get the span tags & replace whatever might be contained rather than a quick text replace, but we should change this before sending off for translation.


jbum (approved):
Awesome sausages.

PR #3243, opened 2021-02-10
2021 02 10/vaccination groups equity link arrow

chachasikes:
Updated arrow based on markup recommendation from Aaron.

Our style docs need some love:

Looks like the content team did use the internal style guide has wrong markup (it's missing a closing div & something is off with its centering)

The display of all the components in WordPress is really good/useful/clear (and serves benefit of testing the integration of our components & our system wide CSS) - 

The interface to get the code snippets from the Wordpress post is a little hard to use - because Wordpress doesn't indent code & the layouts are different. 

The storybook instance doesn't match, so that threw me off:
https://wonderful-plant-07a82e81e.azurestaticapps.net/?path=/story/covid19-arrow-links--light-background (the light blue should be dark blue)

Also noticed it would be nice to make Storybook show cut & paste code snippets, too - there are code snippets, but they are a little different in that they are the Story snippets - which isn't useful to anyone, really.


aaronhans (approved):
Good points about all the doc shortcomings

PR #3246, opened 2021-02-11
Jbum statedash2

aaronhans:
jbum's state dash v2 coming through

The plan is to have this run on its own /v2-state-dashboard/ url as a soft launch starting tomorrow. We will let it replace the live state dashboard late next week if it fares well.

Improvements like making all these charts use the "click to interact" tableau trigger that prevents them from downloading everything right away will begin in the next version. Making the tableau embeds on this page ADA compliant isn't at the top of the tableau dev priority list so it might be faster to get some of these to meet basic accessibility requirements by swapping a few of them out with lightweight javascript ones over the few weeks.


chachasikes (approved):
Great work!

PR #3259, opened 2021-02-11
fix tab links, mismatched div tags

chachasikes:


kkoryaka (approved):
Looks like you are fixing filter button tags

aaronhans (approved):
thanks for troubleshooting this

PR #3263, opened 2021-02-11
Fixed filter markup on sandbox.

jbum:
Fixed filter tags on sandbox to match. I've also repaired the code in WP to match.


chachasikes (approved):

carterm (approved):
now it matches the start tag

kkoryaka (approved):
there you go

PR #3268, opened 2021-02-11
removing old pings

carterm:
These magically appeared after a merge somewhere.  They shouldn't be generated anymore now that we are using PRs for translation updates.

Just casually removing 173 files from the build, no big whoop.


aaronhans (approved):
nice cleanup

PR #3271, opened 2021-02-11
CCG-1480 Adding ethnio code to vaccines

carterm:
Adding a script that only appears for the Spanish version of vaccines page.  Already added to the Wordpress version.

Not that this will generate an empty comment for non-spanish...`<!---->`

### DIFF
--- a/pages/translated-posts/vaccines-es.html
+++ b/pages/translated-posts/vaccines-es.html
@@ -9,7 +9,7 @@ addtositemap: true
 ---
 
 <!--{% if 'lang-es' in tags %}-->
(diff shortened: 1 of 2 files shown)


carterm:
https://digital-ca-gov.atlassian.net/browse/CCG-1480

chachasikes (approved):
ok - let's be away of interactions with future JS charts - should be ok, but heads up

cc @jbum @aaronhans

aaronhans (approved):
This is cool, I like adding these temporary page specific tracking page scripts into the post body, nice that we can restrict it to one lang too

PR #3273, opened 2021-02-11
add chart data label for race and ethnicity chart, reactivate optiona…

chachasikes:
…l data labels in case we need more chart footnotes for the other charts.

Also update Docs a little more.


jbum (approved):
Looks good from Burbank.

aaronhans (approved):
Love these docs

PR #3274, opened 2021-02-11
2021 02 11/vaccinations groups data labels sync

chachasikes:
Update data label text

Sync code from master
Sync vaccines.html markup to this vaccine.html page

Reapply our chart text


chachasikes:
I don't like this one.

PR #3284, opened 2021-02-12
Vaccination by groups

chachasikes:
This is the javascript code for the vaccination by groups chart (with no markup).


aaronhans (approved):
Beautiful

PR #3287, opened 2021-02-12
add some fixes to IE rollup configuration

chachasikes:
A few fixes to the builds for IE

Size & a new scss file
data path

working on issue with es5 version of d3


aaronhans (approved):
looks good!

PR #3289, opened 2021-02-12
manually copy changes from PR ##3288 to avoid issues with class renam…

chachasikes:
…es from production_debug


aaronhans (approved):
nice

PR #3290, opened 2021-02-12
add vaccination by groups

chachasikes:
PR straight to master because preproduction is wrong & being a blocker
this won't work in IE but we will fix that later


aaronhans (approved):
ship it

PR #3311, opened 2021-02-16
Commented out unused vars from home page njk.

jbum:


aaronhans (approved):
Thanks for the cleanup!

Once you are sure these are not used I would remove the lines entirely instead of just commenting them out

PR #3341, opened 2021-02-18
CCG-1542, IE11 fixes for master.

jbum:


chachasikes (commented):
I'll let this one have an extra review - but I see you are adding async & importing v2 dashboard code

kkoryaka (approved):
I guess this is what is needed for IE.

PR #3380, opened 2021-02-22
fixed Scrollable element is not keyboard accessible issue by removing…

kkoryaka:
… scroll property from WP figure


chachasikes (approved):
Looks good

PR #3382, opened 2021-02-22
Task/ccg 1477 fix translation delivery

aaronhans:
This fixes a bug where translated values were not coming through because the references to these dataset fields were broken. There was no problem with the English version because we have English fallbacks


chachasikes (approved):
nice attributes!

PR #3383, opened 2021-02-22
Use staging data for VE charts. Dynamically adjust chart size.

jbum:


chachasikes (approved):
code looks good

kkoryaka (approved):
looks like URLs update

PR #3385, opened 2021-02-22
Task/ccg 1477 fix translation delivery

aaronhans:
This looks good in preproduction, going to production

The dataset fields were not all named correctly and some were failing to pass through translated string values.


carterm (approved):
Is there a corresponding change to the code that uses this?

aaronhans:
No, there is no corresponding change to the code using this. The code that was attempting to use this now works as expected.

PR #3398, opened 2021-02-23
CCG 1615 - Common chart. Code cleanup.

jbum:


aaronhans (approved):
cool

chachasikes (approved):
Cool - You are adding the common chart renderer we discussed. It makes sense to do it this way - cause so far every embed has lots of important different labeling, but the chart drawing behavior stays the same.

PR #3404, opened 2021-02-23
Remove county logic from schools for what's open search

chachasikes:


aaronhans (approved):
nice

carterm (changes requested):
<p>${ac["0"] === "Schools" ? ac["6"] : ac["6"]}</p>

inline statement looks like it always has the same result

carterm (approved):

PR #3409, opened 2021-02-23
use different urls for spanish feedback

aaronhans:
We want different urls on the Spanish pages for the links presented to users after they submit the single field feedback form at the bottom of every page.

- Added new data to the langData file for all languages
- Added a new 11ty filter to get the language specific value
- Called this new filter passing in the field name so we can reuse it for positive and negative survey urls


kkoryaka (approved):
Super!

PR #3416, opened 2021-02-24

aaronhans:
using npx to do an install and then an execution should be avoided in the future...

this hardcoding of a specific package version should fix all the build failures


chachasikes (approved):
Cool. That looks like the recommended fix from the force-resolution issue.

chachasikes:
nevermind

PR #3417, opened 2021-02-24

aaronhans:
This does fix the build problems

But this is a temporary solution, future fix will remove hardcoded package version and hopefully avoid using npx to do the install as well as the execution


chachasikes (approved):
Cool. That looks like the recommended fix from the force-resolution issue.

PR #3419, opened 2021-02-24

aaronhans:
PR to fix the build to master, this fix has been verified in preproduction, the build works and site looks good at preproduction.covid19.ca.gov


chachasikes (approved):
Looks good

PR #3429, opened 2021-02-24
fixed pdf icons displayed inside of the list (ccg-1625)

kkoryaka:


aaronhans (approved):
right on, thanks for addressing this

PR #3431, opened 2021-02-24
CCG-10012 Changed from layerY to offsetY to address differences in Chrome browser that caused glitchy tooltips.

jbum:


kkoryaka (approved):
Seems like moving tooltips around

PR #3441, opened 2021-02-25
added landmark and aria-label to the feedback (ccg-1559)

kkoryaka:
previous PR fixed it partially, need to apply same solution to the buttons too.


aaronhans (approved):
nice

PR #3453, opened 2021-02-25
Show schools guidance

chachasikes:
Schools change to show guidance
Remove a little bit of unused logic

Requires a coordinated content push with Peggy scheduled for later this afternoon.


aaronhans (approved):
good

PR #3506, opened 2021-03-03
removed container's left and right padding when that container is ins…

kkoryaka:
…ide of another container (ccg-1676)


aaronhans (approved):
ok

aaronhans:
I am reverting this change, it fixed the desired bug but had far reaching consequences on mobile layouts perhaps because the author did not anticipate the level of unnecessary nesting being used in many places

PR #3509, opened 2021-03-03
Moved several duplicated routines into src/js/common.

jbum:


aaronhans (approved):
cool

chachasikes (approved):
great!!! yay Jim!

PR #3516, opened 2021-03-03
removed container's padding when that container is inside of another …

kkoryaka:
…container (ccg-1676)


chachasikes (approved):
lookin' good

PR #3534, opened 2021-03-04
CCG-1688 rollover for 3/4

m-sullivan7:
Updating to use today's dates as per Anneli Wong not all vaccination data has been updated.


chachasikes (approved):
looks good

PR #3536, opened 2021-03-04
Added substitution utility that works with both double and single curly brackets. Using this for HPI charts.

jbum:
The intent is to leave in single-curly support, but favor double-curlies, so that we can incrementally fix the substitution strings on wordpress without breaking anything.

### DIFF
--- a/src/js/charts-sandbox/charts/cagov-chart-vaccines-hpi-by-people/index.js
+++ b/src/js/charts-sandbox/charts/cagov-chart-vaccines-hpi-by-people/index.js
@@ -3,7 +3,7 @@ import getTranslations from "./../../../common/get-strings-list.js";
 import getScreenResizeCharts from "./../../../common/get-window-size.js";
 import rtlOverride from "./../../../common/rtl-override.js";
 
-class CAGovVaccinesHPEPeople extends window.HTMLElement {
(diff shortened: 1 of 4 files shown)


chachasikes (approved):
Looks good!

Here are some small code notes:
* can you add spaces before ":reformatReadableDate" so it's clear it's just a hash? 
* applySubstitutions should be fine with this implementation & a known dictionary - adding some error handling to prevent value = undefined or null from rendering & if subdict[key] is not found would be good long term - cause another developer might bumble your dictionary & cause an error.
* maybe rename subdict to something else (can't think of a great alternative... substitutions? I try to avoid abbreviations especially - because it means developers have to learn other developers abbreviations & keep them implemented... which isn't a big deal in a quick turnaround, but over time the weight of that accrues on everyone & it makes the code harder to work with & less desirable to fix because it takes extra energy to load the code in our brains.
* CAGovVaccinesHPEDose has typo declaration & assignment

All the rest looks super great :)

jbum:
Addressed all of Chach's excellent recommendations except error handling (this will fail quietly if symbols are incorrect). Probably best to supply a warning, although I can imagine normal situations where a larger dictionary is supplied than is needed.  Most problematic is presence of an unknown symbol in the source text.

PR #3537, opened 2021-03-04
Swirl/charts centering fix per adam

aaronhans:
The problem reported by Adam is that charts on desktop layouts aren't centered and their footnotes are not the same width.

This CSS change affects all D3 charts but it makes them look slightly better at big screen sizes on /vaccine and /equity and doesn't break the layout on mobile


aaronhans:
not clean PR

jbum (approved):
Looks saner.

PR #3538, opened 2021-03-04
remove max width on charts because width is controlled by parent

aaronhans:
The problem reported by Adam is that charts on desktop layouts aren't centered and their footnotes are not the same width.

This CSS change affects all D3 charts but it makes them look slightly better at big screen sizes on /vaccine and /equity and doesn't break the layout on mobile

### DIFF
--- a/src/css/_charts.scss
+++ b/src/css/_charts.scss
@@ -16,10 +16,6 @@
     display: none;
   }
 
-  .svg-holder {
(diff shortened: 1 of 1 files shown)


chachasikes (approved):
sounds like issues with the parent got resolved - i can test this more once it's up on preproduction

PR #3539, opened 2021-03-04
Curly bracket substitutions and symbol rename.

jbum:


aaronhans (approved):
cool, works both ways

PR #3540, opened 2021-03-04
Swirl/charts centering fix per adam

aaronhans:
Adam reported new charts were not centered or same width as captions below them

Adam reviewed style changes on preproduction and approved


aaronhans:
This change affects equity charts too, those still look ok

jbum (approved):
Another elegant fix from Aaron Hans.

PR #3541, opened 2021-03-05
divide by total combined doses for %

aaronhans:


jbum (approved):
The ratios add up this way. I have no idea what the intention was. :-)

aaronhans:
I am going to merge this to preprod, we'll wait for approval from Angie or Hannah before doing anything to production but based on the chart caption this seems like the right calculation

PR #3543, opened 2021-03-05
Added date hack for HPI charts. 

jbum:
If the latest_administered_date (for HPI chart data) is today, force it to yesterday.  Adds a common utility, getSnowflakeStyleDate, to fetch YYYY-MM-DD strings for today, yesterday or others using a dayDelta.

Removed several old console.logging messages.


aaronhans (approved):
less logs, more dates

PR #3549, opened 2021-03-05
Added date hack for HPI charts. Removed some console logging.

jbum:
The hack prevents "today" from showing up as the last_administered_date for the HPI charts (if it's set to today, we display yesterday).


chachasikes (approved):
great

PR #3559, opened 2021-03-06
Updating March 5 notes

m-sullivan7:
Removing notes about March 5 vaccines administered data from homepage and state dashboard 1.0.

I confirmed data has ticked over on the homepage and state dashboard 1.0.

Thanks for being available on a Saturday Jim!


chachasikes (approved):
looks good

PR #3562, opened 2021-03-07
use new field for first dose column

aaronhans:
Per Hannah:

Back of envelope math shows we have too many data points in the top chart, vaccinations should represent the 7M people vaccinated not the 10M shots given

which generate: cagov/covid-static#265

So the SQL has been updated to remove the dose 2 people from first dose counts here https://github.com/cagov/Cron/pull/106 and this PR updates the frontend to use the new FIRST_DOSE_ONLY field


jbum (approved):
This should cause the first dose numbers to shrink.

aaronhans:
closing because no longer necessary, data changed to make this update

PR #3566, opened 2021-03-08
CCG-1720

m-sullivan7:
Language approved by Hannah and stakeholders (via Angie). Anneli confirmed that Tableau charts were stable over the weekend and note no longer needed on state dashboard 1.0.


chachasikes (approved):
looks good

PR #3569, opened 2021-03-08
added javaScript code that adds icons to external and pdf links (ccg-…

kkoryaka:
…1553)

### DIFF
--- a/src/js/arrow/links.js
+++ b/src/js/arrow/links.js
@@ -1,0 +1,16 @@
+// Adding pdf span to the links with class pdf
+var pdf = '<span class="pdf-link-icon no-underline-icon" aria-hidden="true">PDF</span><span class="sr-only"> (this is a pdf file)</span>';
+
+var pdflLink = document.querySelectorAll(".pdf");
+for (var i=0; i < pdflLink.length; i++) {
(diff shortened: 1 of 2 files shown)


src/js/arrow/links.js
```diff
@@ -0,0 +1,16 @@
+// Adding pdf span to the links with class pdf
+var pdf = '<span class="pdf-link-icon no-underline-icon" aria-hidden="true">PDF</span><span class="sr-only"> (this is a pdf file)</span>';
+
+var pdflLink = document.querySelectorAll(".pdf");
```
aaronhans, line 4: We could make the selector more strict by looking for links with pdf in the href like document.querySelectorAll("a[href*='.pdf']")

chachasikes (approved):
I see that you are adding two new classes (.pdf and .external) - and this will listen for those links, add span tag to update them all.

idea/question: instead of putting this in src/js/arrow/ would src/js/links make sense?

src/js/arrow/links.js
kkoryaka: I see your point. That would stylize all pdf links. But what if content team doesn't need that extra span? This way it kind of give them more control. (Although... automation is is our goal. This is tough one)

PR #3571, opened 2021-03-08
added caption class (ccg-1671) and updated TOC styles (ccg-1528)

kkoryaka:
Moving it live


chachasikes (approved):
This also changes the arrow back to \35 (circle) in the CaGov font.

PR #3572, opened 2021-03-08
added JavaScript code that adds icons to external and pdf links (ccg-1553)

kkoryaka:
Moving it to production


chachasikes (approved):
I see a double L in the pdfLink & externalLink - would be good to clean that up in the next pass (along with putting the code in a /links folder if that makes sense
externallLink

kkoryaka:
@chachasikes you are making a lot of sense. I will create the ticket clean it up in the next pass.

PR #3576, opened 2021-03-09
Change display time from 11am to 10am (19 GMT -> 18 GMT)

jbum:
On homepage,…state-dashboard and v2-state-dashboard.


aaronhans (approved):
thanks for digging into this Jim!

jbum:
This should not be merged until we have our first 10am deployment. This may be Wednesday the 10th.

PR #3577, opened 2021-03-09
Auto Tier Update 03-09-2021 

wpserviceuser:


chachasikes (approved):
This is correct based on our list.

carterm (approved):
Looks legit.  Ready to move when we get the word.

PR #3579, opened 2021-03-09
Ccg 1739/jbum translation leeway

jbum:
Added support for parens in addition to curly braces for dictionary-based symbol substitution for used for D3 charts.

A test will be forthcoming to insure the desired symbols are present.  If parens are used a warning will be issued, but it doesn't need to fail.


aaronhans (approved):
cool, thanks for looking into testing this

PR #3585, opened 2021-03-09
add css for mask effectiveness table

aaronhans:
This CSS is needed for the table displaying the mask effectiveness on /masks-and-ppe/

Opening code review to master and preprod simultaneously, will verify in preprod first


kkoryaka (approved):
Let's roll with it

PR #3586, opened 2021-03-09
add css for mask effectiveness table

aaronhans:
This CSS is needed for the table displaying the mast effectiveness on /masks-and-ppe/


kkoryaka (approved):
Nice looking css

PR #3596, opened 2021-03-10
update accordion to fix a11y issue and fix security vulns

aaronhans:
- Fix security vulnerabilities identified by npm
- Install latest patch version of accordion that includes accessibility fix from Konstantin

verified build still works with subdependency package updates on staging


chachasikes (approved):
nice

PR #3598, opened 2021-03-10
Bug/ccg 1563 a11y tabindex

aaronhans:
Security fixes flagged in npm audit
Accessibility fix identified by Konstantin in bumped accordion


chachasikes (approved):
cool - I'll port this to staging too (if not already) for testing safer economy accordion issue today

PR #3609, opened 2021-03-10
removed external css class since there is already JavaScript method of appending external icons (CCG-1753)

kkoryaka:


aaronhans (approved):
love to see code deletion, nice Konstantin!

jbum (approved):
Looks good!

PR #3613, opened 2021-03-10
Update dashboard.njk

m-sullivan7:
Cases note approved by Los Angeles County director of public health and CDPH.


chachasikes (approved):
Looks good

PR #3621, opened 2021-03-11
Removing March 10 LA county note

m-sullivan7:
To publish after 10:00 a.m. data update. Approved by Angie.


chachasikes (approved):
looks clean

PR #3634, opened 2021-03-11
Ccg 1769/vaccine equity published date

jbum:
Adding date substitutions for vaccines equity charts


carterm (approved):
Seems legit.  PUBLISHED_DATE should be ready now.

PR #3635, opened 2021-03-11
Added table-gridded styles to reopening.scss (CCG-1756)

kkoryaka:


chachasikes (changes requested):
This is different from what I pushed earlier to day - can you pull from staging & update there?
There's a border that Adam noticed was missing & I tweaked the padding/line-height to match Figma.

PR #3643, opened 2021-03-11
update @cagov/accordion component

chachasikes:
sync package.json and package-lock.json from master, update @cagov/accordion


aaronhans (approved):
cool, get staging more inline with master

PR #3644, opened 2021-03-11
Ccg 1575/statedash 2.0 code updates

jbum:
Numerous updates to support new chart on v2-state-dashboard. Updates some of the shared code, styles, and updates the content on v2-state-dashboard.


aaronhans:
I would add this to the full-bleed element at the bottom of this page: margin-bottom: -2.5rem;

maybe add a new utility class for it

Then you don't have to universally shorten the padding on the page container

PR #3648, opened 2021-03-12
Task/ccg 1776 temp add ethnio

aaronhans:
Adding ethnio to homepage temporarily so Clara can survey people


chachasikes (approved):
is the "//... " valid with IE? it might not matter but flagging in case it trips you up later - it's tripped me up with images in the past.

e.g. src="//

chachasikes:
(i know it can be like that if the code needs to run for http & https sites)

PR #3655, opened 2021-03-12
spacing below top charts

aaronhans:
I think this is what we want in order to decrease the spacing between the top 3 charts and the captions below

This has been merged to staging

I haven't modified the WordPress file to add that new class yet


aaronhans:
woops, meant to go to preprod

aaronhans:
updated this to not include the WP HTML, I will add that class directly to WP

jbum (approved):

PR #3656, opened 2021-03-12
Ccg 1575/min height charts state dash

aaronhans:
I think this is what we want in order to decrease the spacing between the top 3 charts and the captions below

This has been merged to staging

I haven't modified the WordPress file to add that new class yet


jbum (approved):
Looks good, thanks Aaron!

PR #3668, opened 2021-03-12
Auto Tier Update 03-12-2021 

carterm:


chachasikes (approved):
This is ready for when we go live

PR #3673, opened 2021-03-12
revert temporary message

chachasikes:


carterm (approved):
removing text and re-adding padding

PR #3674, opened 2021-03-12
Ccg 1575/the great renaming

jbum:
Renaming pages in order to make v2-state-dashboard our new state-dashboard.

Renames the page state-dashboard. v1-state-dashboard

Renames the page v2-state-dashboard -> state-dashboard.

Makes the javascript for v1-state-dashboard at dashboard-v1/ to avoid confusion.

Applies a styling fix.


aaronhans (approved):
nice

PR #3676, opened 2021-03-12
add accordion crawl exclude class on table accordions

chachasikes:
add accordion crawl exclude class on table accordions


aaronhans (approved):
perfect

carterm (approved):
Adding class

PR #3684, opened 2021-03-13
Fix homepage date for new summary boxes on home page.

jbum:


chachasikes (commented):
seems ok - but lets have extra reviews

aaronhans (approved):

PR #3690, opened 2021-03-14
Ccg 1743/daylight savings adjust

jbum:
Note: This will cause times to appear too early (until DST kicks in). 

Should be merged between 2am and 9am on Sunday Mar 14th.

Although our clocks jump forward, we adjust GMT _backward_.  Why?  Right now when it is 6pm GMT, it is 10am here.  After DST kicks in, we roll forward, but GMT does not. Then, when it is 6pm GMT it will be 11am here. So we roll back GMT from 6pm to 5pm to compensate.


chachasikes (approved):
merges changes from master and timezone updates some dates

PR #3691, opened 2021-03-14
Rolled back canned GMT time displays by one hour to account for PST going forward.

jbum:
Note: This will cause times to appear too early (until DST kicks in). 

Should be merged between 2am and 9am on Sunday Mar 14th.

Although our clocks jump forward, we adjust GMT _backward_.  Why?  Right now when it is 6pm GMT, it is 10am here.  After DST kicks in, we roll forward, but GMT does not. Then, when it is 6pm GMT it will be 11am here. So we roll back GMT from 6pm to 5pm to compensate.

### DIFF
--- a/pages/_data/statsData.js
+++ b/pages/_data/statsData.js
@@ -7,7 +7,7 @@ const data = {
     totalDeaths: caseStats[0]['NUMBERDIED'],
     totalDeathsIncrease: caseStats[0]['NUMBERDIED_DAILYPCTCHG'] * 100,
     testsReported: caseStats[0]['TESTED'],
-    lastStatsDate: caseStats[0]['DATE'] + 'T18:00:00Z'
+    lastStatsDate: caseStats[0]['DATE'] + 'T17:00:00Z'
   }
 };
 
--- a/pages/_includes/main.njk
+++ b/pages/_includes/main.njk
@@ -72,7 +72,7 @@ formatNumber(tags,1)-%}
 				<div class="hero-stats" role="region" aria-labelledby="tracking-covid">
 					<h2 class="color-orange text-center m-0 pb-2 pt-3" id="tracking-covid">{{varStatsHeader}}</h2>
 					<div class="text-sm text-center text-300">
-						{{- varUpdateText | replace("[UpdatedDate]", (dailyStatsV2.data.cases.DATE + "T18:00:00Z") | formatDate2(true,tags,+1)) | replace("[DataDate]",dailyStatsV2.data.cases.DATE | formatDate2(false,tags)) -}}
+						{{- varUpdateText | replace("[UpdatedDate]", (dailyStatsV2.data.cases.DATE + "T17:00:00Z") | formatDate2(true,tags,+1)) | replace("[DataDate]",dailyStatsV2.data.cases.DATE | formatDate2(false,tags)) -}}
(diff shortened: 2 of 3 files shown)


chachasikes (commented):
double checking... why are we changing the times of things that run on a cron back an hour...will the services than run not also shift to happen at the same time?
(same comment would apply to the previous commit)

chachasikes:
as you said above "Neither UTC nor GMT ever change for Daylight Saving Time (DST). However, some of the countries that use GMT switch to different time zones during their DST period."

chachasikes:
... and assuming the GMT connects back to our formatDate2 function (which gave us trouble on new years day with some interpreters on the year

'T18:00:00Z' - 

   const locales = 'en-US';
    const timeZone = 'America/Los_Angeles';
  if(datestring&&datestring.indexOf('Z') > -1) {
      const date = new Date(datestring);
      return `${date.toLocaleDateString(locales, { timeZone, day: 'numeric', month: 'long', year: 'numeric' })} at ${date.toLocaleTimeString(locales, { timeZone, hour: 'numeric', minute: 'numeric' })}`;
    } else if(datestring === 'today') {
      const date = new Date();
      return `${date.toLocaleDateString(locales, { timeZone, day: 'numeric', month: 'long', year: 'numeric' })} at ${date.toLocaleTimeString(locales, { timeZone, hour: 'numeric', minute: 'numeric' })}`;
    } else {
      return datestring;
    }
  });
  
  
  This does adjust for locales - so making sure we take this into account @jbum

jbum:
Hi Chach,  The cron services that are supposed to run by 10am (all the crons actually) have times specified in GMT. We are rolling those GMT times back as well in a separate PR that has already been merged.  So the services will also be matching (I believe the data that those cron jobs are reading is prepared well in advance).

This particular patch affects a few "last updated" times that are displayed on the Home Page and State Dashboard pages, among others. For some of these pages, we want to say "10am" regardless of whether it's daylight savings or not (and we want the crons to run at the same local time).  Right now (in PST) 10am pacific corresponds to 6pm GMT (which we specify in the code as 18:00 GMT). However, after DST takes effect, 10am pacific will be 17:00 GMT.  At 18:00 GMT, we will have "sprung ahead" to 11am.

The formatDate2 function converts a passed in GMT time to "America/Los Angeles".  Right now it will convert 18:00 to "10am" but tomorrow it will convert 18:00 to "11am" so the spots where we say "18:00" to force a display time of "10am" need to be adjusted.

There is probably a better long term fix in which we specifically request "10am in Pacific Time" without passing in a GMT value. But this is the simplest fix that will prevent stuff from breaking in the morning.

I may be totally wrong, of course, which is why I'll be there to see what actually happens (and quickly correct, if need be).

abquirarte:
Thank you all for being vigilant and taking care of our site!
Call me if anything comes up. 🙏🏼

~ Angie, from my cell.

jbum:
Wow Chach, I stand corrected. Totally missed the "PST" in dateFormat2.  Gonna close this PR without merging it (although we still need to roll the crons back I think - we will check 'em later this morning).  However, this creates a new issue.  Now the date offset we use in these three places uses PST while the crons use the current timezone, I'm pretty sure, which is confusing. Don' t need to fix this now, however.

PR #3704, opened 2021-03-15
update yolo link

aaronhans:
bad link for Yolo alerts


aaronhans:
This is an old page, the code on it is not even a web component, that data shouldn't live in the site but don't know if it is worth refactoring

chachasikes (approved):

PR #3715, opened 2021-03-15
Patch reopening matrix to use "en" reopening matrix table

chachasikes:
so that the accordions will at least not show an error (the content on that matrix was required with this release) - will revert when translation issues are resolved.
 
( also push recent changes from master to preproduction)


kkoryaka (approved):
Nice looking patch

PR #3716, opened 2021-03-15
Roll back times an hour for DST.

jbum:
Production is currently showing 11am on home page, state-dash and 3pm on equity.

I believe we want 10am and 2pm, respectively. This fixes that.


chachasikes (approved):
Notes on this since dates, daylight savings & timezones are notoriously confusing:
* Cron was reverted ahead manually an hour relative to GMT - corresponding manual date conversions are updated here.
* We have some weird code for when no time zone is found (if no timezone is found) - Jim added a comment about this.

PR #3733, opened 2021-03-15
Patch reopening matrix to use en table so that the accordions will at…

chachasikes:
Patch reopening matrix to use en table so that the accordions will at least be correct & not show an error (the content on that matrix & required with this release) - will revert when translation issues are resolved.

We will revert this once the translations come through but can't tell where we landed on that publish so prepping this PR for now.


jbum (approved):
Localization is good.

PR #3734, opened 2021-03-16
update default state for accordions for accessibility

aaronhans:
The default state for accordions should have the -1 tabindex


aaronhans:
verified in preproduction

chachasikes (approved):
happy tabbing!

PR #3735, opened 2021-03-16
update default state for accordions for accessibility

aaronhans:
The default state for accordions should have the -1 tabindex


chachasikes (approved):
awesome catch

kkoryaka (commented):
Cool

PR #3744, opened 2021-03-16
2021 03 16/fix translations homepage

chachasikes:
Fix homepage updates section so it can use the translated files.


aaronhans (approved):
thanks for fixing this Chach!

Sorry for the slow review, I got confused by all the files

chachasikes:
I experimentally added the master sync (which I was avoiding for this reason - but I think we have to do syncs in separate merge requests even if this would be easier)

PR #3745, opened 2021-03-16
CCG-1788 Corrected admin date display adjustment to work for old data.

jbum:
Insures that the 'data from (administered)' date is always at least one-day behind the published date.

Does not use 'today' 'yesterday' logic anymore.


aaronhans (approved):
great

PR #3747, opened 2021-03-16
CCG-1809. Equity update date time fix.

jbum:
Changes update date on equity page to 10am. With next data pull, this should show up as Mar 17 (Wed) 10am.


jbum:
Note: This requires that the date in the data be pegged to "Tuesday". That is the case this week. However, the last couple week's it's been set to Monday.  Will need to keep an eye on this, and edit the data if necessary.

aaronhans (approved):
cool

PR #3750, opened 2021-03-17
remove this survey invite

aaronhans:
remove the original covid19 survey from the site at Clara and JP's request


jbum (approved):
Appears to remove a survey-related script.

PR #3758, opened 2021-03-17
Translation - schools-may-reopen-in-these-counties

jorgv:


chachasikes (approved):
Looks clean

chachasikes (changes requested):
wait... the date format changed this might be bad.

chachasikes (approved):
Ran this & checked safer economy & state dashboard - these variables are display only & not used for calculations.

chachasikes:
@jbum this looks fine - but the tier data is also used in state dashboard - looks like it's used same way was for safer economy & page rendered fine - but adding you - we run these other dates through formateDate2 function

aaronhans (approved):

PR #3759, opened 2021-03-17
Translation - vaccines

jorgv:


chachasikes (approved):
Looks clean - about to stage next translation on top of this one

PR #3762, opened 2021-03-17
Quick fix for HPI chart cropping issue.

jbum:


jbum:
This fixes the immediate issue for Angie.  A better fix will be forthcoming that keeps the bars out of the legend area.

aaronhans (approved):

PR #3766, opened 2021-03-17
Forcing right-alignment for state-wide line caption on risk-factor chart.

jbum:
The old code attempted to move the legend out of the way of the bars, but was not very effective. At the moment, the data wants the legend to be on the right. This may change. Started a discussion about improvements to chart design on Jira with Adam.


chachasikes (approved):
thumsbup

PR #3771, opened 2021-03-17
Translation - vaccines

jorgv:


chachasikes (approved):
conflict resolutions looks clean

PR #3775, opened 2021-03-17
CCG-1086/equity margin issues

jbum:
Repair ragged margins on some sections of equity page, per Adam's feedback.


chachasikes (approved):
looks good

PR #3777, opened 2021-03-17
remove this survey invite

aaronhans:
Removing basic covid survey per Clara and JP. The vaccines survey remains


chachasikes (approved):
bye survey

PR #3786, opened 2021-03-18
Ccg 1819/new hpi data source

jbum:
Using new data source for HPI charts.  Merging this will cause preproduction to use the /v2/ json file.


chachasikes (approved):
cool, staged data references

PR #3787, opened 2021-03-18
CCG-1819/new hpi data source

jbum:
Warning: Merging this will cause production to use the /v2/ json file.  Fire when we are actually ready.


chachasikes (approved):
also looks good & same

aaronhans (approved):
thanks Jim!

PR #3792, opened 2021-03-18
Ccg 1829/old pages cleanup

jbum:
Removing old translated versions of pages that no longer exist.


aaronhans (approved):
nice

aaronhans (approved):
this is already redirected, thanks for cleaning files up

PR #3803, opened 2021-03-19
CCG-1823 Added colors and hatch-fills to HPI charts on vaccines page.

jbum:
@aaronhans note: I'm noticing these are breaking on IE11.  The prior charts are breaking as well. 

We'll need to repair (and possibly have an alt-solution if IE11 doesn't support SVG patterns).


chachasikes (approved):
wanna see these hatches when the build is done! :)

jbum:
Hatches are visible on staging.
Note: HPI charts are currently broken on IE11 (old versions too).

PR #3805, opened 2021-03-19
CCG-1086/equity margin issues

jbum:
Gets several margins in line with each other.


chachasikes (approved):
if preprod looks good then this looks good

chachasikes (commented):
@jbum I see two divs added at the end - is that the fix?

chachasikes:
The div issue doesn't seem to be an issue

PR #3816, opened 2021-03-22
Task/ccg 1835 npm upgrade

aaronhans:
This is the big upgrade to lockFile version 2 for npm v7

It also includes all latest dependabotprs (11ty was bumped with security fixes in dependencies which removed our requirement to do subdependency overrides)

If this works properly the team will need to all switch to npm v7 together


aaronhans:
There are a lot of commit changes here from dependabot but this is all related to package upgrades so only files changed are package and package-lock

These updates have already been deployed to staging

aaronhans:
there is an actual issue running tests, looking into it

aaronhans:
chucked the image shrinking libraries: I think we should do stuff like that outside of this repo in the future, there was also a weird issue with express after all the upgrades so I chucked it in favor of web-dev-server in a spawned process and jest with forced exit

chachasikes:
I'll pull this code & try to run the update & tell you results with my current npm & node versions.

chachasikes:
This looks good

Procedure:
1. Go to the local covid19 directory. `cd path/covid19` with the lastest checked out.
2. Remove node modules folder. `rm -rf node_modules`
3. Make sure you have node version `v14.15.3` and npm version: `6.14.0` (This npm version comes with node version:  `v14.15.3` )
4. Install packages like normal. `npm install`
5. Run build, `npm start` It should run gulp services like normal and succeed.

chachasikes (approved):
Looks good. Nice work.

I added a comment detailing my procedure for the install.

PR #3818, opened 2021-03-22
Ccg 1726/wordy language wrap

jbum:
Caption wrapping for HPI chart for Spanish and Tagalog.

If caption is longer than X characters, it displays as two lines using word-count to split.


aaronhans (approved):
pretty cool

PR #3819, opened 2021-03-22
removed PDF icon from arrow link (ccg-1853)

kkoryaka:


carterm (approved):
changing the way the pdf icon renders

chachasikes (approved):
suppressing the icon (in arrow links I believe)

PR #3820, opened 2021-03-22
Ccg 1823/glorious colors

jbum:
Color changes on vaccines/hpi and equity/health-places charts.


kkoryaka (approved):
I see white and blue instead of light blue and yellow. I guess it must be glorious combination.

PR #3821, opened 2021-03-22
Task/ccg 1835 npm upgrade

aaronhans:
This is the big update to package-lock to move to lockFile version 2

During this upgrade I also:

- merged in all outstanding dependabotprs (resulting in a long commit history)
- removed the image compression packages because we should be doing image processing in covid-static
- replaced express with web-dev-server in the test script because it was having problems running after the upgrade


aaronhans:
This update has been merged to preproduction and staging

aaronhans:
force resolutions stuff also dropped, no longer necessary after recent 11ty update

chachasikes (approved):
okay let's do it! @aaronhans

PR #3823, opened 2021-03-22
Fixed HPE charts for IE11.

jbum:
Config needed an update so charts wouldn't break in IE11.  Also added explicit styling. HPE Charts look a bit different, but at least they are showing up now.


jbum:
This addresses CCG-1867

aaronhans (approved):
thanks for checking on IE

PR #3825, opened 2021-03-22
Ccg 1823/glorious colors.  Color charts.  IE11 fix.  Caption fix.

jbum:


aaronhans (approved):
nice, not leaving IE11 behind

PR #3842, opened 2021-03-23
CCG-1808. Forcing 1 decimal place on several charts that were permitting 0 (but usually using 1).

jbum:
Based on a discussion with Michael and Hannah.


kkoryaka (approved):
0 to 1, seems like a good move.

PR #3860, opened 2021-03-24
CCG-1808 Added a decimal place to deaths numbers.

jbum:
Wordpress fixed to match.


kkoryaka (approved):
2 is better than 1 I guess

PR #3897, opened 2021-03-29
Task/ccg 1727 new homepage section

aaronhans:
New homepage section which has 3 columns on desktop to feature 3 pages, currently vaccines, what's open and tier map. There is a graphic and text link for each. The tier map svg file is built dynamically by a data service to apply the css colors to the separate svg file. svg files returned from the covid19 host are currently being gzipped.

The content for this section was moved into WordPress as a fragment so it will be translated.

translated-post placeholders are created so the build works while we wait for translated content


chachasikes (approved):
big change!
adding snippets to all the translated homepages

git ignoring the cali counties colors files (could that eventually be named re: its meaning - like california-county-tier-map.svg ?)

super cool use of fills and colors to make the map (adding to my mental map of reusable map components needs)

styling changes

PR #3898, opened 2021-03-29
Task/ccg 1727 new homepage section

aaronhans:
New homepage section 

working in preproduction, ready to go to production


chachasikes:
looked at it in a few devices on preproduction looking good

FUTURE FEATURE - discuss the black border - our county tier map doesn't have a black border, so it creates a design inconsistency @aaronhans 

reviewing this code now...

chachasikes (approved):
Seems right

PR #3907, opened 2021-03-30
bump package for go-to-top to get slightly longer delay. Made a note …

chachasikes:
…in the new package to let this be a data-attribute, but that will require some good data checks


chachasikes:
Staging maxed out - merging anyways

PR #3918, opened 2021-03-31
Correcting day-adjustment for equity missingness chart.

jbum:
The date provided with this data is sometimes Monday, and sometimes Tuesday. Previously used a fixed 1 day offset to make it align with the Tuesday dates on the rest of the page. Currently this is forcing it to Wednesday, which is undesirable.  

Now looking at the actual weekday provided and computing a dynamic adjustment to Tuesday for any of Sunday,Monday,Tuesday.  If Wed-Sat, we leave it as it.


aaronhans (approved):
cool

PR #3919, opened 2021-04-01
Task/ccg 1746 update links under stats

aaronhans:
updated the set of links under the top box numbers on the homepage with Konstantin's code, added translation fixes and style fixes to top call to action on homepage

These updates have been reviewed by editorial and design and verified on preproduction


chachasikes (approved):
Looks good & ran changes locally since there were lots to verify it builds.

PR #3922, opened 2021-04-01
making sure that accordion titles are left aligned (ccg-1822)

kkoryaka:


chachasikes (approved):
Q: will this work with RTL? If not we might need to look at that with an addition all class to support right alignment.

PR #3927, opened 2021-04-01
2021 04 01/roadmap airtable v1

chachasikes:
Create a staging area for testing syncing data and small data UX changes to What's Open in advance of additional changes to What's Open search. 

Content team can export a "snippet" of code from Airtable into Wordpress syntax, so that the data can be managed in Airtable & aligned with our existing table data in advance of moving other parts. 

* Airtable script "snippet maker" script lives in src/js/roadmap-airtable-v1.
* Alternative v1 data source (whats-open-airtable-v1) table data json is mapped and loaded into this component.
* New data source also has Peggy's preferred table header names to help ensure changes are correct.
* What's open alternate component uses that new table structure
* Rollups added to support this new component.

To port this to preproduction the plan is to:
* Train Airtable edits & test copying the snippet to Wordpress. (Note JSON alternates will come later but we are focusing on Airtable training right now, so that's a very low priority)
* Publish Airtable data to CURRENT data source
* Sync code changes from v1 alternate to original roadmap code (test first on staging once content team is comfortable with the change)
* Tests on staging, then re-test on preproduction
* Then go live.

This is an "as-is" change with the only changes being how we publish & the table header name change & corresponding references in code.

Once we get through that we will be able to pass through other upcoming changes.


aaronhans (approved):
cool, good plans

PR #3928, opened 2021-04-01
IE fixes

aaronhans:
- fix IE homepage readability by removing css that was blocking the whole background image
- get babel polyfilling working again by reducing market share check
- remove unnecessary npx install
- add defensive check into state dash which messes up IE because all the files are combined and that is outside of a web component

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
   },
   "scripts": {
     "devserver": "web-dev-server --root-dir docs/",
-    "preinstall": "npx npm-force-resolutions@0.0.3",
     "dev": "cross-env NODE_ENV=development npm run start",
     "start": "npm run watch",
     "serve": "npx @11ty/eleventy --serve --quiet",
--- a/src/css/_hero-stats.scss
+++ b/src/css/_hero-stats.scss
@@ -119,9 +119,6 @@ hero-stats {
 		@include absolute(left 0 top 0);
 		@include size(100%);
 		z-index: -1;
-		@include ie {
-			height: auto !important;
-		}
 		&-pic {
 			@include size(100%, 70vh);
 			@include absolute(left 0 top 0);
--- a/src/js/es5.js
+++ b/src/js/es5.js
@@ -2,7 +2,7 @@ import './config/staging.js';
 import '@webcomponents/webcomponentsjs';
 import 'whatwg-fetch';
 import './polyfills/endswith.js';
-import '@cagov/accordion';
+// import '@cagov/accordion'; accordion indicator svgs are broken, disable accordions in IE until fixed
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -60,7 +60,6 @@
   },
   "scripts": {
     "devserver": "web-dev-server --root-dir docs/",
-    "preinstall": "npx npm-force-resolutions@0.0.3",
```
aaronhans, line 63: no longer necessary since 11ty fixed the subdependency security issues, we had already removed the list of packages to force prior to this so this step already did nothing

src/css/_hero-stats.scss
```diff
@@ -119,9 +119,6 @@ hero-stats {
 		@include absolute(left 0 top 0);
 		@include size(100%);
 		z-index: -1;
-		@include ie {
```
aaronhans, line 122: this css was breaking homepage readability in IE by blocking display of both background image and color causing the white h1 to be displayed on white background

src/js/es5.js
```diff
@@ -2,7 +2,7 @@ import './config/staging.js';
 import '@webcomponents/webcomponentsjs';
 import 'whatwg-fetch';
 import './polyfills/endswith.js';
-import '@cagov/accordion';
+// import '@cagov/accordion'; accordion indicator svgs are broken, disable accordions in IE until fixed
```
aaronhans, line 5: commenting out accordion stuff in the IE only js file because the svg open/close indicator isn't working in IE.

right now in production all accordions are off due to the js errors so this does not change the state of production, we should fix the svgs and turn accordions back on for IE though

when accordions are not working they show up fully expanded so all content is still readable

src/js/es5.js
```diff
@@ -18,11 +18,11 @@ import './vaccines/charts/ie11.scss';
 import './charts-sandbox/index.js';
 import './dashboard-v2/index.js';
 
-import applyAccordionListeners from './tracking-you/index.js';
-// twitter widget doesn't support IE11 so not including here
-window.onload = (event) => {
-  applyAccordionListeners();
-};
+// import applyAccordionListeners from './tracking-you/index.js';
```
aaronhans, line 21: more accordion stuff removed in the IE file

src/js/rollup.config.es5.js
```diff
@@ -36,7 +36,7 @@ export default {
           {
             modules: false,
             targets: {
-              browsers: '> 1%'
+              browsers: '> 0.5%'
```
aaronhans, line 39: IE now has less than 1% market share so we have to bump this down to get critical automatic polyfilling

chachasikes (approved):
great IE changes

PR #3929, opened 2021-04-01
IE fixes

aaronhans:
- fix IE homepage readability by removing css that was blocking the whole background image
- get babel polyfilling working again by reducing market share check
- remove unnecessary npx install
- add defensive check into state dash which messes up IE because all the files are combined and that is outside of a web component

### DIFF
--- a/src/css/_hero-stats.scss
+++ b/src/css/_hero-stats.scss
@@ -119,9 +119,6 @@ hero-stats {
 		@include absolute(left 0 top 0);
 		@include size(100%);
 		z-index: -1;
-		@include ie {
-			height: auto !important;
-		}
 		&-pic {
 			@include size(100%, 70vh);
 			@include absolute(left 0 top 0);
--- a/src/js/es5.js
+++ b/src/js/es5.js
@@ -2,7 +2,7 @@ import './config/staging.js';
 import '@webcomponents/webcomponentsjs';
 import 'whatwg-fetch';
 import './polyfills/endswith.js';
-import '@cagov/accordion';
+// import '@cagov/accordion'; accordion indicator svgs are broken, disable accordions in IE until fixed
 import '@cagov/step-list';
 import './dashboard-v2/async-polyfill.js';
 import './alerts/index.js';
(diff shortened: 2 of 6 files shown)


src/css/_hero-stats.scss
```diff
@@ -119,9 +119,6 @@ hero-stats {
 		@include absolute(left 0 top 0);
 		@include size(100%);
 		z-index: -1;
-		@include ie {
```
aaronhans, line 122: this was totally breaking the homepage background layout in IE causing the main H1 to be unreadable

src/js/es5.js
```diff
@@ -2,7 +2,7 @@ import './config/staging.js';
 import '@webcomponents/webcomponentsjs';
 import 'whatwg-fetch';
 import './polyfills/endswith.js';
-import '@cagov/accordion';
+// import '@cagov/accordion'; accordion indicator svgs are broken, disable accordions in IE until fixed
```
aaronhans, line 5: currently in production all accordions are inactive in IE because of many js errors preventing execution, luckily we use progressive enhancement so everything is in an expanded state, commenting this out does not change the state of production systems but we should fix the svg indicator because accordions did work nicely in IE

src/js/es5.js
```diff
@@ -18,11 +18,11 @@ import './vaccines/charts/ie11.scss';
 import './charts-sandbox/index.js';
 import './dashboard-v2/index.js';
 
-import applyAccordionListeners from './tracking-you/index.js';
-// twitter widget doesn't support IE11 so not including here
-window.onload = (event) => {
-  applyAccordionListeners();
-};
+// import applyAccordionListeners from './tracking-you/index.js';
```
aaronhans, line 21: also accordion related, this is the IE only file

src/js/rollup.config.es5.js
```diff
@@ -36,7 +36,7 @@ export default {
           {
             modules: false,
             targets: {
-              browsers: '> 1%'
+              browsers: '> 0.5%'
```
aaronhans, line 39: This is the big problem that was breaking everything in IE

carterm (approved):
Excellent comments.  Removing npx.  Checking for a non-null divElement before using it.  Removing cagov accordion usage.  Changing browser module target.

src/js/rollup.config.es5.js
chachasikes: interesting

chachasikes (commented):
Trying to review... posting a comment to see if my review gets unstuck

PR #3937, opened 2021-04-01
CCG-700/ten charts

jbum:
Chart work, in progress.  These charts are rendered on the charts-sandbox page (only visible locally and on staging).  I'd like to merge this code into master (and then preprod -> staging) to make it less cumbersome to update staging.

Charts are all rendered by a common routine: histogram.js

Still need to add 4 more charts hidden by tabs, tool-tips, county-display, and tab support.


aaronhans (commented):
10 charts!

aaronhans (changes requested):
little worried about the package-lock changes here, let's remove that file if we didn't intentionally make changes

jbum:
@aaronhans package-lock removed (I checked in the one from master). Not sure what happened there.  Tooltips added, can be previewed on staging.

PR #3958, opened 2021-04-05
Task/ccg 1953 move labels to wp

aaronhans:
Deleted translation strings from the files in the repo, moved them into WordPress post, code uses WP strings, translations have been provided by AvantPage


chachasikes (approved):
nice change

chachasikes:
(Fails are related to max number build envs)

PR #3971, opened 2021-04-06
2021 04 05/go to top length

chachasikes:
Bumping version of go-to-top on staging.

This version accepts attributes from html tag, and connects them to common-page-labels so that the go-to-top duration can be changed at will while we continue to test the duration that feels best.


chachasikes:
@aaronhans sneak peek.

Still testing mysterious package issues locally before I merge this but as you can see the change is supposed to be simple.

chachasikes:
This was buggy & broke the build, so I reverted it on staging.

Needs more investigation, something is off/wrong.

PR #3980, opened 2021-04-06
Auto Tier Update 04-06-2021 

wpserviceuser:


chachasikes (approved):
HOLD

These values are confirmed to match expectation.

PR #3984, opened 2021-04-06
Update with the self-hosted agent

SaraAhmad944:
Replacing the MS hosted agent with the self hosted agents.


aaronhans (approved):
changes from CDT to use self hosted agent for build pipeline

PR #3987, opened 2021-04-06
2021 04 06/preproduction go to top

chachasikes:
Bump go-to-top feature, works with Wordpress data & translated now.


jbum (approved):
+1

chachasikes:
holding on this - but prepped in case staging goes well

PR #3988, opened 2021-04-06
bump go-to-top & connect it to WP data

chachasikes:
Bump go-to-top feature, works with Wordpress data & translated now.


chachasikes:
holding on this until staging PR works correctly, but prepped ahead of time

PR #4017, opened 2021-04-07
Translation - equity

jorgv:


carterm:
newer translation deployed

PR #4045, opened 2021-04-08
updated spacing for dark accordion first item, removed tabindex from accordion container in eleventy.js (ccg-1563)

kkoryaka:


chachasikes (approved):
Nice

PR #4051, opened 2021-04-09
unknown value for age is wrong coming from service, hide it until we fix

aaronhans:


aaronhans:
The value for unknown is coming back wrong from the SQL the service is running, hide value until we figure out why

PR #4053, opened 2021-04-09
bump accordion to development version of v2.0.9 (which is a PR to main)

chachasikes:


chachasikes:
@kkoryaka  @aaronhans  

I cleaned up the v2.0.9 to prep it for publishing.
* Clean up comments
* Make comments generally sentences
* Updated Readme & a story to match the original screenshot
* Added index.scss in the STORY (not the component) to style the demo content (a table).


This packaging option seems to be working again.

I used https://gitpkg.now.sh
to point to the PR to main on Storybook.

These were the settings.

Screen Shot 2021-04-08 at 7 07 15 PM

Then, I updated staging with the url to this pre-published package.
& That is what we will test the the staging build now.

PR #4074, opened 2021-04-12
applied styles to the native WP separator (ccg-1863)

kkoryaka:


aaronhans (approved):
nice new class, like the viewportwidth usage here

PR #4075, opened 2021-04-12
Trying to fix white space by unsetting the height and width and giving .sr-only class text-indent property instead (ccg-1892)

kkoryaka:
…

### DIFF
--- a/src/css/_general.scss
+++ b/src/css/_general.scss
@@ -2424,3 +2424,9 @@ font {
   }
 }
 
+/* Fixing White space */
+.sr-only {
+  text-indent: -10000px;
+  height: unset !important;
+  width: unset !important;
+}


chachasikes (commented):
Nice!
Looks good on staging. 

No one is gonna find that giant text-indent.
I see that there is another place where .sr-only has `absolute !important` 
It is set by bootstrap. That absolute makes this safe & not pushing around any relative divs.

If we port this to a component or something, we'll need to retain the absolute position.

For general knowledge: https://developer.mozilla.org/en-US/docs/Web/CSS/unset
"The unset CSS keyword resets a property to its inherited value if the property naturally inherits from its parent, and to its initial value if not. In other words, it behaves like the inherit keyword in the first case, when the property is an inherited property, and like the initial keyword in the second case, when the property is a non-inherited property."

PR #4081, opened 2021-04-13
removing vaccines ethnio code

carterm:
Clara requested we remove the enthio script from the vaccines page.

Also removed from WP source.


chachasikes (approved):
Looks good

PR #4095, opened 2021-04-13
fixing white space by unsetting the height and width and giving .sr-only class text-indent property (ccg-1892)

kkoryaka:
…


chachasikes (approved):
YAY!

PR #4110, opened 2021-04-14
Fixed number formatting bug in equity chart.

jbum:
Missing parameter to formatting function was causing labels to not appear in tabbed charts on the bottom of the equity page.


aaronhans (approved):
thanks Jim!

PR #4111, opened 2021-04-14
Fixed number formatting bug in equity chart.

jbum:


aaronhans (approved):
thanks for finding this

PR #4126, opened 2021-04-15
update links on homepage

aaronhans:
verified in preproduction
- use hardcoded link to myturn for external link (can change this to variable with localized filter when link points internally again)
- update links under top stat boxes to use localize url filter so they stay in translated pages


chachasikes (approved):
preprod links looked good & this looks the same

PR #4130, opened 2021-04-16
Added new list styles (ccg-2042)

kkoryaka:


chachasikes (approved):
(the table that I saw on staging looks nice)

PR #4167, opened 2021-04-19
2021 04 19/safer economy tiers

chachasikes:
https://cadotgov.slack.com/archives/CVCSH3CR5/p1618435717450400?thread_ts=1618435547.449800&cid=CVCSH3CR5 (from this thread)

* Add tier based search that's integrated with new table data for safer-economy-lang.
* One feature not enabled yet is a line between the tier data thats combined with the county search - still looking into the hooks for Awesomeplete, if we can connect to the input event & apply a class, there is a class called .separator we can use. Doing this with straight up CSS doesn't work because the list filter results are dynamic & it's not readily possible to add classes to the data.

### DIFF
--- a/pages/_data/pubData.js
+++ b/pages/_data/pubData.js
@@ -70,7 +70,7 @@ const files = [
         'activityLabel'
     ]},
     Table2: {
-      pivot: true,
+      pivot: false,
       require: [
         'id',
         'reverseValue',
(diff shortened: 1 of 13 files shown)


chachasikes:
data-county-tiers="{{pubData['en'].saferEconomyLang.Table2 | dump | escape}} is added to safer-economy.html & needs to be stored in WP.

chachasikes:
https://digital-ca-gov.atlassian.net/browse/CCG-2067

PR #4168, opened 2021-04-19
copied state dash translation files over so new AvantPage translation…

aaronhans:
…s can be accepted

I have copied the translated posts from state-dashboard and created files for v3-state-dashboard. Files have to exist before translations can be accepted. This will allow the translate menu to work and we can add the translate tag to the v3 post so that AvantPage translates it


jbum (approved):
Translated files copied over, thanks!

PR #4203, opened 2021-04-21
Task/ccg 2064 chart drawer

aaronhans:
CSS updates to support accordion variant for chart information


chachasikes (approved):
looks good

notes: 
* ".accordion-alpha" seems extra branded to include "alpha" is it necessary (it is inherited from accordion?)
* the cagov-plus and cagov-minus are dependencies. we need to communicate dependencies & make sure they are imported. 
* i'll add cagov-plus and cagov-minus to our list of components

aaronhans:
thanks, good points

PR #4258, opened 2021-04-28
Ccg 700/final fixes

jbum:
Fixed to bugs that Daniel found.

Inhibit 'Pending' from showing on reported-date graphs.

Show County chart titles with the word "County" - added separate strings for these, and temporarily working around lack of translations for these, supporting both the old and new formats.

Live on staging.


aaronhans (approved):
cool

PR #4262, opened 2021-04-29
Date fix for date at top of equity page.

jbum:
Forces date at top of equity page to Wednesday.  Total hack.


chachasikes (approved):
Looks like a date change.

Not sure why it's Wednesday specific or what's broken, but looks like you are setting data publishing to today, (which makes sense.)

jbum:
Not happy with this method. Gonna tweak it.

PR #4271, opened 2021-04-29
Video section translation support P1

carterm:
Makes the json data translation compatible.  Will need to modify the "Active" column to be blank or non-blank and then submit for translation.  Once translation is confirmed, we can switch the page to use the new files.

Addresses...


chachasikes (approved):
Translations for video content

PR #4280, opened 2021-04-29
Removed v3 from most translated titles.

jbum:
Leaving Chinese alone because I can't make out if it's in there.


kkoryaka (approved):
I see some added titles

PR #4290, opened 2021-04-30
Activating video translations

carterm:
Translations came in and look great!  This update will make it so the homepage uses them now.

(Also did some whitespace formatting for the affected section)

Addresses...


aaronhans (approved):
thanks for noticing and fixing this!

PR #4291, opened 2021-04-30
"From the Governor’s Office" date translations

carterm:
Translating the dates/times on the news section on the homepage, even though we can't translate the titles.


aaronhans (approved):
nice

PR #4293, opened 2021-05-01
fixing reversed phone numbers for AR emotion support page

carterm:
using non-breaking-hyphens to fix a few phone numbers that didn't reverse.
Updated the Wordpress page to prevent it on future releases.

Addresses...


aaronhans (approved):
nice

carterm:
Going to scrap this one since AvantPage is going to handle these numbers differently with explicit LTR.

PR #4297, opened 2021-05-03
Add ID to pagerating section so that it can be linked to

daguar:
Per @aaronhans's direction, this PR adds an id to the page rating section so that it can be linked to directly (which we want to do on the in-development broadband benefit page)


aaronhans (approved):
Nice!

daguar:
Thank you @aaronhans !

PR #4299, opened 2021-05-03
update proxy to one in prod subscription group

aaronhans:
Use the new proxy. The new proxy is a brand new Azure function which is in the production subscription group and is updated to fix a sub-dependency security vulnerability


kkoryaka (approved):
Seems like changing search URL azure group

PR #4300, opened 2021-05-03
delete old workflow

aaronhans:
We don't need this anymore


kkoryaka (approved):
yml workflow file is deleted

PR #4353, opened 2021-05-06
adding server redirect for 410

carterm:
Adds server code to properly display a 410 gone page.


aaronhans (approved):
cool

nice that new page slug is being created completely in wordpress

PR #4389, opened 2021-05-11
fixed accordion cut off bug in search results (ccg-2172)

kkoryaka:


carterm (approved):
Adding a media condition

PR #4467, opened 2021-05-18
Added support for #location-county anchors on state-dash and equity page.

jbum:
When county or statewide is selected, the URL gets updated with

#location-<county>
or
#location-california

When the state-dashboard page is loaded, if the URL contains #location-<county> (and the county name is valid) it performs a county search. Same with equity page.  This enables specific counties to be bookmarked and shared more easily.  Does not work on Vaccines page.


aaronhans (approved):
very cool Jim!

PR #4468, opened 2021-05-19
Bug/equity dash bug

aaronhans:
Dr Vargo noticed production bug where this line doesn't go away.

This is caused by the legend using the same class
The scale of the chart changes in one of the states so then we see two versions of the line

Fix is to add a unique class selector to the full line so we can remove it instead of the line in the legend


aaronhans:
Screen Shot 2021-05-18 at 6 43 24 PM

PR #4475, opened 2021-05-19
Fixed GA error tracker.

jbum:
1. Switched to send to our main GA account so we can monitor it.
2. Turned off code which depends on preloading of error events, which we are currently not doing.
3. Fixed syntax of ga() so it works properly.
4. Added error.message to the label.


aaronhans (approved):
very nice, thanks for fixing!

PR #4477, opened 2021-05-20
CCG-2074 Added filename/linenumber to event action desc.

jbum:
CCG-2074. These fields will often contain no information, but hopefully we'll see a few from certain browsers.


aaronhans (approved):
cool

PR #4486, opened 2021-05-20
fixed dark accordion text contrast issue (ccg-2296)

kkoryaka:


jbum (approved):
Thanks so much for figuring this out!

PR #4504, opened 2021-05-21
Adjust spacing for safer economy page

chachasikes:
Add additional spacing for Safer Economy page, highlight box.


jbum (approved):
Elbow room.

PR #4541, opened 2021-05-27
add tracking code for vax ctas

aaronhans:
Added a tracking section which looks for data attributes and applies GA event parameters.

These event tracking values were defined by Britt, details in the ticket


carterm (approved):
Adding JS block

aaronhans:
The data attributes are applied to the https://staging.covid19.ca.gov/vax-for-the-win page

PR #4542, opened 2021-05-27
translation placeholders

carterm:
Adding placeholders for translations in English just so we don't get broken links whilst we wait.


aaronhans (approved):
thanks Carter!

carterm:
Tested this locally

PR #4565, opened 2021-05-28
new zip method

aaronhans:
update from suresh


carterm (approved):
changed to linux command

PR #4630, opened 2021-06-07
Removed tags from aria-labels on three chart types.

jbum:
Screen-reader fixes for three chart styles on the equity page. Formerly, the aria-labels duplicated the tooltip markup, which contains tags. This strips out the tags, condenses spaces and removes leading/trailing space in the aria-labels.


chachasikes (approved):
Looks good & like you are trimming and doing some other stuff to the caption. 

Notes:
* A comment would be helpful so we don't have to look up the regex expressions.
* caption.trim() might work too.

PR #4631, opened 2021-06-07
CCG-2361 Removed tags from aria-labels on three chart types.

jbum:
Screen-reader fixes for three chart styles on the equity page. Formerly, the aria-labels duplicated the tooltip markup, which contains tags. This strips out the tags, condenses spaces and removes leading/trailing space in the aria-labels.


aaronhans (approved):
nice

PR #4645, opened 2021-06-08
CCG-2371 Added support for statewide pop numbers.

jbum:
This backward compatible update shows state-wide population numbers and legends if the data is present in the supplied JSON data.  If the data are not supplied, the charts are the same as before.  The vaccines-by-groups charts currently on staging show these additions.


aaronhans (approved):
cool

PR #4652, opened 2021-06-09
CCG-2372 Removed excess divs to fix tooltip styling.

jbum:


aaronhans (approved):
less divs!

lot of changes in here, assuming just synchs from master to preprod before opening PR

PR #4653, opened 2021-06-09
CCG-2372 Removed excess divs from tooltips to fix styling.

jbum:


aaronhans (approved):
thanks for identifying and fixing this by removing unnecessary markup!

PR #4669, opened 2021-06-10
CCG-2379/tooltips added to vaccines group charts

jbum:
Tooltip display added to vaccines group charts.
Inhibiting display of baseline indicators on exceptional categories (other, unknown, unknown/undifferentiated)


aaronhans (approved):
more tooltips, nice!

PR #4670, opened 2021-06-10
CCG-2379/tooltips added to vaccines group charts

jbum:
I'll merge this after I get sign-off from Angie.


aaronhans (approved):
cool, thanks Jim!

PR #4712, opened 2021-06-15
Removed some code referencing old safer-economy features. Removed dashboard-v1

jbum:


aaronhans (approved):
cool

aaronhans:
you can delete lines instead of commenting them out when retiring stuff

jbum:
I commented in this case, because I'm still convinced we're gonna have another virus outbreak, and those particular site features are useful...

PR #4713, opened 2021-06-15
Changed default text weight to 300.

jbum:
Something recently caused the default body text to change from weight of 300 (correct) to 500 (incorrect).  This overrides the default body weight which was set to 500 -> 300.

This fixes the problem, however, this file (_general.scss) has not been changed in weeks, and the problem is new, so there was likely a different css font-weight override in place that recently disappeared.  Worth looking into...  However, Adam has confirmed that this fix looks good.


aaronhans (approved):
thanks Jim!

PR #4714, opened 2021-06-15
CCG-2403 Added support for date display on homepage summary.

jbum:


jbum:
This fix also removes the time display, which is overly specific (and inaccurate).

aaronhans (approved):
good

PR #4715, opened 2021-06-15
Stop build on template errors

carterm:
This will prevent template errors from deploying


jbum (approved):
Nice! I’m curious if the exception contains any useful info that might be good to relay (e.g. line numbers, files)

carterm:
> Nice! I’m curious if the exception contains any useful info that might be good to relay (e.g. line numbers, files)

All I could find is an exit code.  Whatever errors happen are listed in the console anyway.

PR #4764, opened 2021-06-19
Restructuring of social media icons in footer.

jbum:
Not sure this is a good idea, posting this PR to get feedback from @aaronhans and/or @kkoryaka or others on the engineering team.

The problem that is being solved: In the Covid19 footer, there are three columns. (links, logo/social media icons, hotline info).  In non-mobile views, these columns are side-by-side, and the social media icons are positioned directly under the california logo.  

image

In the mobile version, these columns are displayed one after another, with the social media icons at the bottom (beneath the content of column 3).  

image

For this reason, the social media icons were implemented LAST, outside the columns, and then relative positioning was used in desktop views to push them up snug against the california logo.

The problem with that solution is that the relative positioning changes depending on the longest column (which is typically the links on the left) so as links are added, removed, the relative positioning needs to be changed to compensate for this.  As the links were edited this week, the layout of the social media icons kept breaking.  I wanted to restructure this so that adding/removing links would not move the social media icons on the desktop view.

My solution was to add a _second_ set of social media links for use on the desktop, at the bottom of column 2.  CSS code is used to toggle between the mobile version (outside the columns) and the desktop version (at the end of column 2).  This eliminates the need to have a repositioning offset.  The toggling is accomplished using visibility: collapse / visibility: visible (I couldn't use display:none, because there are sub-elements with display overrides).

I'm aware that I could have probably used a CSS Grid to change the layout (changing the structure of the grid for mobile/desktop), but since we were already using using a flex-grid for the columns, this seemed like a more straight-forward solution.  However, I'm not 100% (and also not sure about browser support for collapse and would appreciate feedback.


aaronhans (approved):
This is good, thanks for finding a solution that will be easier to maintain, I wasn't aware of visibility:collapse but it has great browser support.

css grid is supported by 99% of our visitors so feel free to use in the future if it helps reduce duplicate markup and the non grid fallback is still usable

jbum:
Thanks Aaron, I'm gonna test this a bit more just to make sure there aren't any situations where both icons show up.

PR #4767, opened 2021-06-19
CCG-2441/ie11-hero-image

jbum:
This fixes a bug.  When we recently changed to the beach hero image on the home page, it stopped displaying on IE11.  This happened because we introduced a new feature: a differently cropped version of the image for mobile/portrait display.  

The image switch was accomplished by using CSS code to swap in a different image, changing the 'src' attribute of the image tag using the _content:url_ css property. Since CSS was controlling the image, I removed the original 'src' attribute from the markup.  This broke IE11, because IE11 doesn't like that newfangled _content:url_ CSS property.

I worked around this issue by re-introducing the 'src' tag in the markup (defaulting to the .jpg image, which IE11 can read) (after several failed attempts to fix by just changing the CSS). On more modern browses, the src still gets overruled by the CSS, however this does introduce a performance penalty: _both_ the jpg and webp versions of the image are loaded. 
 I can see it will be more efficient to just use the jpg.  So I'm removing the CSS source that loads it for now.

Tested on staging using Browser Stack.

### DIFF
--- a/pages/_includes/main.njk
+++ b/pages/_includes/main.njk
@@ -98,7 +98,7 @@ formatNumber(tags,1)-%}
 			</div>
 			<div class="hero-photo" role="complementary" aria-labelledby="hero-background">
 				<div class="hero-photo-fade"></div>
-				<img class="hero-photo-pic" alt="Photo Description">
+				<img class="hero-photo-pic" src="/img/hero_v2.jpg" alt="Photo Description">
 				<div id="hero-background" class="hero-photo-description">Photo Description</div>
 			</div>
 		</section>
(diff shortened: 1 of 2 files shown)


aaronhans:
Will using srcset help us serve the desired resolution to everybody efficiently and support all browsers?

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images 

Will the picture approach mentioned at the end of that article simplify our webp insertion too?

aaronhans (approved):
approving to fix IE11 no image, thanks for looking into further optimizations Jim

PR #4779, opened 2021-06-21
CCG-2441/Hero Image responsive code / performance

jbum:
Using responsive code ideas from https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images so we get more optimal performance without breaking IE11.

Reduced size of desktop jpeg from 2000 -> 1600.

Removed some unused images.


aaronhans (approved):
very cool, love how this lets us remove css rules too

thanks Jim!

PR #4832, opened 2021-06-30
Added code to suppress daily totals on summary boxes in expected weekend/holiday conditions.

jbum:
This adds an 11ty filter which is used to check if we need to display a hyphen (em-dash actually) instead of digits for the daily totals.  Currently we want to do this on weekends and holidays.  The conditions I am checking are:

* Is the daily tests number zero? (This is the largest of the three numbers, and should never be zero normally). 

* Is it Saturday or Sunday?

* Is it July 5 or Labor Day?

* Is the difference between the tests.date and meta.published_date greater than one day? (is the data stale?)

Any of these conditions will cause hyphens to display.  

Note I also need to add a block of markup to the state-dashboard page, but can't add it until this is merged into master.

Once we see what actually happens on Saturday, I will pare these conditions down to just the necessary ones.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -167,6 +167,33 @@ module.exports = function(eleventyConfig) {
       return null;
     }
   );
+  
+  /*
+  Return TRUE if we need to suppress the daily totals in the summary boxes (displaying hyphens instead of numbers).
+  This is currently extra conservative, we will remove some of these conditions when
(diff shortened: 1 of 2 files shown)


abquirarte:
@mediajunkie FYI

aaronhans (approved):
cool, thanks Jim

PR #4922, opened 2021-07-22
Removed forced underline. Removed extraneous <span> tag.

jbum:
Fixes a cosmetic issue Peggy noticed. Adam's preferred link style is to remove unlinks on hover. On these links, the underline was being forced, and it was causing a weird issue in which the underline on the space in front of the links disappeared on hover.  This fixes it to behave using Adam's preferred style, in which the underlines disappear completely on hover.

Accomplished by removing explicit "underline" classes in the markup.  Also spotted and removed an extraneous <span> tag.


jbum:
image

aaronhans (approved):
nice

Issue #4992, opened 2021-07-31
GA record click events on in page links


aaronhans:
Link clicks occuring on in page anchor links don't seem to be recorded.

We have TOC sections at the top of pages on covid. We should record click events on these type of events so we can determine how often people follow those links

mediajunkie:
@traintestbritt what is the best way to request this setup?

traintestbritt:
Hi Xian, this way works! Can you provide more context, please? Certain
click events are usually captured so I need to know what clicks and what
page. Also, when is this needed?

traintestbritt:
@mediajunkie your comment came through via email and it didn't show me the whole thread, but now that I am here I have all of the context. I am still misunderstanding exactly what needs to be recorded, unfortunately. I'd want to double-check that @thepegisin's ask isn't already captured in existing analytics and just needs to be surfaced. Can someone explain to me what needs to be captured, please? Perhaps a screenshot example will help clarify for me. Thank you!

aaronhans:
Creating a table of content with links to sections of the page is a common practice on covid19 site now. This is also used on cannabis. On cannabis we have the links in the left sidebar instead. Here is a screenshot of a set of on page anchor links on the covid site.

<img width="967" alt="Screen Shot 2021-08-02 at 5 29 11 PM" src="https://user-images.githubusercontent.com/353360/127939628-12ed3276-0f2a-43c7-bf1b-adb403cc6ad8.png">

Since these links go to a named anchor link farther down the already loaded page there is no page load occurring and I don't see any other existing GA events firing when I follow in page links like this.

I was going to add logic to our event tracking script on covid to capture these clicks as custom events just like we do with offsite and pdf clicks.

I think we also need to do this on the cannabis site.

mediajunkie:
Thanks for that context, Aaron!

traintestbritt:
Ahh, I understand now! Thank you for clarifying @aaronhans. You're right, we have no way of tracking TOC clicks. Please let me know if you need me to review or test once tracking is set up. Thanks again.

aaronhans:
PR merged, events being tracked in GA, screenshot from realtime report below, data will show up in regular event reports starting tomorrow. Thanks for pointing this out so we could instrument it @thepegisin!

<img width="747" alt="Screen Shot 2021-08-04 at 10 20 12 AM" src="https://user-images.githubusercontent.com/353360/128225727-3e88a814-914e-4ccb-81e3-dc5ef38d8301.png">

[...]

PR #5007, opened 2021-08-03
add anchor link tracking

aaronhans:
This adds custom event tracking for anchor links. The following information would be sent (the port :8000 on the eventLabel won't be there in production):

event-labels


traintestbritt (commented):
looks good to me, ty Aaron!

aaronhans:
related ticket: https://github.com/cagov/covid19/issues/4992

Issue #5039, opened 2021-08-06
Prototype charts by vax status


mediajunkie:
(child of https://github.com/cagov/covid19/issues/4982)

- [figma](https://www.figma.com/file/LrzsOu8U5KcMAjJTQ1O3BG/covid19.ca.gov-screens?node-id=7861%3A38915)
- [one pager (not always current)](https://docs.google.com/document/d/1OGVRkTSgRy-BGWhC4HZHsc5iRKzL2oCY9UD1y5VbwGU/edit)

![cases by vax mock](https://user-images.githubusercontent.com/3227378/128549373-9c808644-5d5b-4a27-918d-d1e05777e39b.png)

- [x] line charts
- [x] showing **cases** by default
    - [x] allowing a toggle to show **hospitalizations** and **deaths**
    - [ ] (or possibly just **cases** if that decision happens)
- [x] with data reported per week (https://preproduction.covid19.ca.gov/state-dashboard/)
    - [x] or daily, with a 7-day running average (https://preproduction.covid19.ca.gov/state-dashboard/?weeks=16&mode=daily)
- [x] for a trailing number of weeks (probably 16 or 20 weeks)
- [x] Cases show in terms of number per 100k
- [x] Hosps. and deaths to be shown in terms of number per million
- [x] Data that is substantially incomplete (because too recent) to be indicated as such in the chart design. This lag is as follows:
    - [x] cases at 1 week
    - [x] hospitalizations at 2 weeks
    - [x] deaths at 4 week

![image](https://user-images.githubusercontent.com/3227378/128548998-f2bd1c2b-5b88-4990-82c6-efd2f2247e71.png)

PR #5099, opened 2021-08-12
#5039 v4 state dashboard

jbum:
Implementation of postvax charts for slug /v4-state-dashboard/

Publishing my current code for this slug, so that content team can make edits.

Code has a lot of prototype-stuff in it, but does not affect the live state-dashboard.


aaronhans (approved):
nice!

PR #5134, opened 2021-08-19
Work-around for Safari 14.1 refresh issue with large-tab buttons.

jbum:
There is a safari 14 bug which causes our big-blue (large-tab) buttons to not refresh properly, even though the styling is correct.

This work-around is based on a discussion here:

https://bugs.webkit.org/show_bug.cgi?id=225354

Tested to work in Safari 14.1.


aaronhans (approved):
good one Jim

PR #5186, opened 2021-08-30
Added support for displaying 1 or 2 additional decimal places to displayed counts in equity 100k chart.

jbum:
This adds a decimal place to chart values and tooltip display if the max value falls below 20. It adds a second decimal place if the max value falls below 2.  Currently affects the statewide deaths chart only, by adding a decimal place, since the max value is currently at 10.  May also affect some county-wide charts (although extremely low numbers tend to cause data hiding in these charts, and I haven't found any affected counties yet).

image


aaronhans (approved):
nice improvements

PR #5187, opened 2021-08-31
Added pagesectionincludes shortcode to simplify handling of multiple pages with similar names.

jbum:
Removed some redundant substitution handling for v#-state-dashboard variant versions in nunjucks templates.  

Thank you to Aaron for pointing me to the thing that needed changing.


aaronhans (approved):
nice

PR #5203, opened 2021-09-01
change chart scales

aaronhans:
Bug:
Screen Shot 2021-09-01 at 2 57 49 PM

Fix: 
Screen Shot 2021-09-01 at 2 57 25 PM

Code change:

This code change forks the histogram file just for hospitalizations and uses a version where the domain passed into the bars uses the line values so they match the average line and the y axes scale.

This code change has been pushed to staging.


chachasikes (approved):
Porting/moving the histogram to a dashboard & fixed link to registering it.

PR #5248, opened 2021-09-09
Renamed src/js/dashboard-v3 to src/js/dashboard

jbum:
Organizational fix.

This removes the -v3 from the directory name used for state-dashboard javascript.

We are moving well past v3, and now typically used a shared source directory for future versions.


aaronhans (approved):
nice cleanup

PR #5251, opened 2021-09-09
Tweaks to "Pending" logic on sparklines

jbum:
Using "UNCERTAINTY_PERIOD" fields for computing pending days (to remove) from sparklines, when available (rather than forcing a fixed 1 or 3 weeks).  This way the sparklines match the underlying charts better.

Added an override for this feature, used on the vaccines sparkline, for which no such data is available.

Switched from weeks to days for the units.

Changed tests sparkline to show posivitiy rate instead of #4566 

Swicthed from dynamic to precomputed values for cases/deaths


aaronhans (approved):
nice

PR #5269, opened 2021-09-11
Puppeteer svg gen

aaronhans:


jbum (approved):
Thanks for the late breaking fix!

PR #5296, opened 2021-09-15
Implemented region fill for tests sparkline.

jbum:
Added styling on sparkline renderer (note: it currently includes a circular reference to generated css file)

Added custom testing chart mode "solid-region" which does a region fill using d3.area()

Copied some changes made manually to generated/sparklines.css into the permanent versions of these files, so we don't lose those changes (color changes, opacity).


aaronhans (approved):
cool new chart styles

PR #5315, opened 2021-09-17
Reorganized sparkline rendering code. 

jbum:
* Moved sparkline rendering code out of src/js/dashboard and into src/js/state-dashboard-sparklines/, the only page that actually does d3 rendering of the sparklines.

* Eliminated circular, and now unnecessary css file include and styling from state-dashboard-sparklines.html (will be mirrored on wordpress).


aaronhans (approved):
nice reorg

PR #5344, opened 2021-09-22
Renaming 'state-dashboard-sparklines' to more generic 'chart-renderer'

jbum:
* chart-renderer compliments chart-tester and handles more than just state-dash, and more than just sparklines.
* Will be changing Wordpress stub to match
* Changing covid-static job to reference this new directory name.


aaronhans (approved):
cool rename

PR #5367, opened 2021-09-23
Preloading of cases postvax chart via embedded <img> tag.

jbum:
Candidate solution for making postvax chart visible in IE11, prerenders the chart as an embedded .

Styling isn't quite there yet. And the HTML elements rendered by template are missing (legends).

Fwiw, I still don't understand why ALL the other dynamic svgs work...  Would love to fix dynamic rendering first...


jbum:
Pulling this, not a fan of this technique.

Issue #5363, opened 2021-09-23
Simple Table


zakiya:
Notes:

- [Tables here] https://cannabis.ca.gov/applicants/application-license-fees/ are coming from the caweb state template styles which are minified and pasted into the CAWeb theme tools. 

- We should create a table component in the design system

- [Staging](https://staging.covid19.ca.gov/variants) `do-not-crawl`

- [Figma](https://www.figma.com/file/LrzsOu8U5KcMAjJTQ1O3BG/COVID19.ca.gov-screens?node-id=7727%3A38518)

zakiya:
@m-sullivan7 You can view the new default table styles at https://staging.covid19.ca.gov/variants

m-sullivan7:
Thank you @zakiya ! Is there a way to make the table border clearer? I see it, but it might be too faint for some people.

In the absence of @adam-little (who I'm tagging so he can give an opinion when he returns), I'm going to solicit opinions from the content team on whether we should center the table instead of have it left-justified. Do you know if the content team can do that already? (I'm happy to experiment in WordPress, but didn't want to jump in and disturb anything.)

zakiya:
@m-sullivan7 Here it is with a black border (same as the figma mockup)
![CleanShot 2021-09-24 at 09 39 36](https://user-images.githubusercontent.com/915352/134710720-3912ec29-2556-43b2-84cc-8a9797bcafbd.png)

Here it is with something in between (#b1b1b1)

![CleanShot 2021-09-24 at 09 42 17](https://user-images.githubusercontent.com/915352/134711054-5b715220-ae9a-4a6e-ae32-a2d439c1c5bf.png)

Looks like there's a toggle in the editor for alignment but it isn't functional. I'll work on adding that functionality and post when it's ready for review.

m-sullivan7:
Thanks for the options Zakiya! We're going to go with the first option (black, in line with @adam-little 's Figma). Even if that was only a provisional design from Adam, we're more confident following that.

Thank you for the investigation of the alignment. We would indeed like to center the table if possible, so it will be useful!

zakiya:
@m-sullivan7 Border colors updated and you should be able to change the alignment using the alignment button in Wordpress.
![CleanShot 2021-09-24 at 11 11 49](https://user-images.githubusercontent.com/915352/134721509-599083b9-a27e-4d01-bd73-66158bf3e7b2.png)

m-sullivan7:
[...]

PR #5386, opened 2021-09-27
#5363 Wordpress table styles.

zakiya:


zakiya:
@kkoryaka That ul is nested in .wp-block-table so it'll only affect uls with a Gutenberg table block.

jbum (approved):

kkoryaka (approved):

PR #5392, opened 2021-09-27
Aligned titles on tracker boxes for home page.

jbum:


jbum:
Before
image

After
image

PR #5396, opened 2021-09-28
Pull new vaccines daily average number.

jbum:
WAIT FOR JIM TO MERGE THIS - currently scheduled for the 9:30 rollout Wed Sep 29

NOTE: A similar fix needs to be applied to wordpress (state-dashboard.html) when this is merged.


aaronhans (approved):
changing reference to daily doses average

PR #5574, opened 2021-11-02
Task 4979/menu ribbon

zakiya:


jbum (approved):
Spiffy new menu for large window sizes.

PR #5582, opened 2021-11-03
Adjustments for coming out of Daylight Savings.

jbum:
_Not to be merged_ until Sunday Evening, Nov 7. (Jim'll take care of it).

This patch advances some displayed times which are expressed in GMT by one hour, to account for the increased offset between Pacific Time and GMT after DST is no longer in effect.


aaronhans (approved):
Thanks Jim!

Issue #5555, opened 2021-10-28
Health equity: chart accordion width


m-sullivan7:
I noticed that one chart information accordion is the width of the paragraph text above the chart, rather than the chart it serves. All the other chart information accordions are the same width as the chart.

![image](https://user-images.githubusercontent.com/69169190/139287236-0f36281b-08a5-415a-bad5-b1e9f002d181.png)

When possible, can we change this chart information accordion to be the same width as the chart?

zakiya:
@m-sullivan7 @mediajunkie Proposed changes to equity page here:

m-sullivan7:
Looks great to me. Thank you @zakiya !

(I even spot checked a few other things this time around and found nothing. 😉)

zakiya:
This is merged.

PR #5583, opened 2021-11-03
#5555 Equity page line chart width.

zakiya:
Margin tweak for "COVID-19 health equity metric" at https://covid19.ca.gov/equity/


jbum (approved):
Simplified layout.  Nice!

PR #5584, opened 2021-11-03
Moved references to img files and data files to use Cloudfront hosted data.covid19.ca.gov

jbum:
This is currently being tested on preproduction. It only affects links to files formerly in these two locations:

files.covid19.ca.gov/data/
files.covid19.ca.gov/img/

PDF files and other files in files.covid19.ca.gov should be unaffected.


aaronhans (approved):
nice Jim!

PR #5612, opened 2021-11-09
Equity data sources patch

jbum:
Inserting a data-sources reference into the templates for two charts.

These charts currently have a chart-drawer as part of the template. This inserts some needed markup between the bottom of the chart and the chart-drawer. If the data-source item has not yet been provided (as it won't be for translated pages at first), it is inhibited by using a empty string.

Currently in preview on staging.


aaronhans (approved):
cool

PR #5615, opened 2021-11-10
Using original directory for vaccine equity jsons.

jbum:
Part of a general effort affecting /covid19/ /Cron/ and /covid-static-data/ to consolidate vaccine equity code.

When we switched the eligible vaccine population to 5+, I created a V2 version of the CovidVaccineEquity job to run in parallel to the original. I've now removed original (v1) and gone to a single version of the job using the v2 code.

The /v2/ files have been copied to their original location (the parent directory), and the cron now writes to the original location.

This patch modifies the chart renderer to retrieve those json files from the original (parent) directory, instead of /v2/.


aaronhans (approved):
cook

PR #5646, opened 2021-11-15
cagov/covid19#5635 Break sparklines to two columns on medium screens.

### LINKED ISSUE #5635: Fix alignment on Tracker boxes.
Fix alignment on Tracker boxes at intermediate window sizes (roughly 760 - 980px).  Currently looks like this:

image

* Discuss with Xian and see if we want to switch to a 2x2 display for this range of sizes, which would simplify alignment issues.
* Daily Average and Test Positivity should share a baseline.

zakiya:
Fixes #5635 on homepage.


aaronhans (approved):
nicely done

PR #5656, opened 2021-11-16
Changing files.covid19 references to data.covid19

jbum:
This affects /img/ and /data/ files only.

Currently being tested in preproduction.


chachasikes (approved):
Nice new subdomain.

aaronhans (approved):
Thanks for all your work on this Jim. Glad we are confident and ready to roll this out nwo

PR #5687, opened 2021-11-19
Using static.covid19 for images.

jbum:
Temporarily using static.covid19.ca.gov for images, by way of an 11ty replacement, and for the sparklines.

This will be reverted back to using data next week, after we get the domain name repointed.

Currently in preview on preproduction.


aaronhans (approved):
nice

PR #5713, opened 2021-11-30
Initial workflow for PR static preview.

jbum:
@xjensen take a close look and see if I missed anything.  Big potential for cut + paste errors here....

### DIFF
--- a/.github/workflows/eleventy_build_pr.yml
+++ b/.github/workflows/eleventy_build_pr.yml
@@ -35,12 +35,8 @@ jobs:
       - name: Build 11ty
         run: |
           mkdir dist
-          npm ci --production
+          npm install
           DOMAIN=${URLSAFE_BRANCH_NAME}.pr.covid19.ca.gov npm run build
-      - name: Write robots.txt
-        run: |
-          echo 'User-agent: *' > docs/robots.txt
-          echo 'Disallow: /' >> docs/robots.txt
       - name: Deploy to S3
         uses: jakejarvis/s3-sync-action@v0.5.1
         with:
@@ -55,7 +51,7 @@ jobs:
       - name: Invalidate CloudFront cache
         uses: chetan/invalidate-cloudfront-action@v1.3
         env:
-          DISTRIBUTION: 'E2EGQV7PEACS1R'
+          DISTRIBUTION: 'E3JRVVFYGDA1AM'
           PATHS: '/*'
           AWS_REGION: 'us-west-1'
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/eleventy_build_pr.yml
+++ b/.github/workflows/eleventy_build_pr.yml
@@ -37,6 +37,10 @@ jobs:
           mkdir dist
           npm install
           DOMAIN=${URLSAFE_BRANCH_NAME}.pr.covid19.ca.gov npm run build
+      - name: Write robots.txt
+        run: |
+          echo 'User-agent: *' > docs/robots.txt
+          echo 'Disallow: /' >> docs/robots.txt
       - name: Deploy to S3
         uses: jakejarvis/s3-sync-action@v0.5.1
         with:
--- a/.github/workflows/eleventy_build_pr.yml
+++ b/.github/workflows/eleventy_build_pr.yml
@@ -1,0 +1,70 @@
+name: Eleventy Build PR Site
+on:
+  pull_request:
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/eleventy_build_pr.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Escape branch name for URL
+        shell: bash
+        run: echo "URLSAFE_BRANCH_NAME=$(echo ${BRANCH_NAME} | sed 's|[^A-Za-z0-9-]|-|g' | sed -E 's|-*([A-Za-z0-9]*.*[A-Za-z0-9]+)-*|\1|')" >> $GITHUB_ENV
+      - name: Report escaped branch name
+        shell: bash
+        run: echo ${URLSAFE_BRANCH_NAME}
+      - name: Build 11ty
+        run: |
+          mkdir dist
+          npm ci --production
```
xjensen, line 38: Looks like this `npm ci --production` line might be failing, perhaps because there's no `package-lock.json` in the repo. You could replace this with `npm install`. I'd recommend it, for now.

.github/workflows/eleventy_build_pr.yml
```diff
@@ -0,0 +1,70 @@
...
+          DOMAIN=${URLSAFE_BRANCH_NAME}.pr.covid19.ca.gov npm run build
+      - name: Write robots.txt
+        run: |
+          echo 'User-agent: *' > docs/robots.txt
+          echo 'Disallow: /' >> docs/robots.txt
```
xjensen, line 43: I'm not sure what this robots.txt job will do here, since Covid already writes a default `robots.txt` from the `rootcopy` folder. Something to watch.

.github/workflows/eleventy_build_pr.yml
```diff
@@ -0,0 +1,70 @@
...
+          AWS_S3_BUCKET: 'pr.covid19.ca.gov'
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-west-1'   # optional: defaults to us-east-1
+          SOURCE_DIR: ./docs # only move built directory
+          DEST_DIR: pr/${URLSAFE_BRANCH_NAME}
+      - name: Invalidate CloudFront cache
+        uses: chetan/invalidate-cloudfront-action@v1.3
+        env:
+          DISTRIBUTION: 'E2EGQV7PEACS1R'
```
xjensen, line 58: Recommend changing the value here to match Covid's CloudFront distribution for PR sites. It's "E3JRVVFYGDA1AM".

.github/workflows/eleventy_build_pr.yml
jbum: Oh yeah, we should nuke that bit...

xjensen:
@jbum It works!

As expected, the production robots.txt file snuck through into the PR site. 

https://jbum-5692-pr-preview-sites-workflow.pr.covid19.ca.gov/robots.txt

We'll need to bring back the not-production robots.txt file in some way. 

On Drought, we have a separate section in the production GitHub Action to write a different robots.txt file, instead of doing a copy from the `rootcopy` folder. [Here's the code](https://github.com/cagov/drought.ca.gov/blob/3458a2046bdddeddf280581c1ebb3e061489e5cf/.github/workflows/eleventy_build.yml#L28-L32). That's one option. There are probably other ways to do it too. Let me know if any questions about this.

aaronhans:
Excellent, nice job @jbum! I love this feature @xjensen!

Thanks for pointing out the nocrawl robots logic too Jon

jbum:
OK, will work on the robots.txt issue shortly...

jbum:
@xjensen The code sample you pointed at for drought creates an 'enable all' robots.txt file.  I assume the purpose of this override is to make a 'disable all' robots.txt file for the PR preview site, so I did that, using this code:

```
      - name: Write robots.txt
        run: |
          echo 'User-agent: *' > docs/robots.txt
          echo 'Disallow: /' >> docs/robots.txt
```

If I'm missing something, let me know...

xjensen:
Yes, that works.

https://jbum-5692-pr-preview-sites-workflow.pr.covid19.ca.gov/robots.txt

My main concern was whether or not that "Disallow" robots.txt step in the GitHub Action could overwrite the default robots.txt file that's copied over from `pages/rootcopy`. Looks like it does. So we're good to go!

Apologies for being unclear about this.

PR #5725, opened 2021-12-01
Dashboard Date Corrections

jbum:
_DO NOT MERGE_ until approval from Xian.

_REQUIRES_ an accompanying patch to the wordpress post.

Prior to this patch, the production site shows more inaccurate times on the dashboard pages.  The time on the state-dashboard page and vaccination-progress page was forced to 10:00AM, and the date AND time on the equity page was forced to the most recent Wednesday at 10:00AM.  

This patches correct that situation.  A separate workflow in covid-static-data keeps track of update dates-times to key JSON files used on the dashboard pages, and they are stored in two files in the CDN.

https://github.com/cagov/covid-static-data/tree/main/data/status

This patch uses the data in these files to display more accurate update dates and times.  The time on the top of the dashboard pages is rounded to the next half-hour mark, and a more accurate update time is displayed below the tracker boxes.  Samples:

image
image
image

Note that these times are collected at build time, to improve site performance.  Production will need to be rebuilt to display the most accurate times.  This is normal for regular site updates, but late-breaking updates will require a site-rebuild.

NOTE: Also using this PR to test Jon's PR static preview system.


pages/_data/equityDashboardUpdateTime.js
```diff
@@ -0,0 +1,20 @@
+// equitydashboardUpdateTime
+//
+// Retrieves json file for use in state dashboard
+//
+const fetch = require('node-fetch')
+
+module.exports = function() {
+  let dataDomain = 'https://raw.githubusercontent.com/cagov/covid-static-data/main/';
```
aaronhans, line 8: Can we get this from cloudfront or is that possibly not up to date in this case?

aaronhans (approved):

jbum:
Current preview when dates differ

image

pages/_data/equityDashboardUpdateTime.js
jbum: Fixed.  Was thinking by reading from github during builds, we can slightly speed up rollout, but probably not necessary at this point.

jbum: Changed back. Let's discuss later -- I'm using github both to speed the pipeline, and so I don't have to do a series of cloudfront updates/cache expiries during the morning update.  I'll work on a longer term fix that does a cloudfront update around 9:55 to insure sync for CDPH.

jbum:
Closing this, will merge directly.

PR #5732, opened 2021-12-03
Support image urls in both files and static locations.

jbum:
Fixes findlinkstolocalize() to support images in either files.covid19 or static.covid19 locations, so that translated images show up as designed.

Temporary fix until we get files.covid19 domain repointed.


aaronhans (approved):
nice

PR #5748, opened 2021-12-06
new workflows to write to S3, invalidate cloudfront

aaronhans:
This adds deploy steps to the production and preproduction branches deploy to S3 and then invalidate the CloudFront cache of the associated distribution.

The preproduction workflow has been run successfully after changes in that branch


xjensen (approved):
Looking good.

PR #5754, opened 2021-12-07
replace deploy to gh-pages branch

aaronhans:
The azure deploy process is watching the gh-pages branch for changes which it then grabs for its deploy workflows

This PR adds the new S3 and CloudFront steps to covid but also does not remove the gh-pages deploy section which is critical for deploying to azure but adds a comment explaining what that step is for


jbum (approved):
Restores crucial section with a useful comment.
⭐️

PR #5756, opened 2021-12-07
Working redirect code

carterm:
shows how to capture the redirect data from the wordpress table.

This code reads the `page-redirect-table` in covid and creates a table that can update s3.  This will account for 301 and 410.  This uses the same method cannabis is using.  There is a code sample to update the bucket wide redirect config If manual wild card redirect systems for translations etc need to be manually added.


aaronhans:
Excellent, thanks Carter!

carterm:
I have not tested it live, but there is a debug launch in the project to test the creation of the dataset to upload to S3.

carterm:
The code works, but it has not been added to the deployment script.

aaronhans:
I added integration into production build workflow

Also removed old Azure preproduction workflow because that migration to AWS is complete

aaronhans (approved):

jbum (approved):
Thanks Aaron!

PR #5757, opened 2021-12-08
Added workflow support for auto-building staging on Tuesdays at 1:40pm

jbum:
This very simple patch is intended to autobuild staging at 1:40pm on Tuesdays for equity preview purposes.

This works by reusing the 'staging env build and deploy' workflow, aka

azure-static-web-apps-wonderful-ground-0c34d501e.yml

I added a workflow_call trigger so it can be called from another workflow, and then a separate workflow which triggers it like a Cron job.

* Note that if/when Staging gets moved to S3, any other staging builds will also need to be triggered in the same way.

* If this works, I may remove the morning auto builder system and do a similar trigger when Covid-static-data does a merge on the dashboard file.

### DIFF
--- a/.github/workflows/azure-static-web-apps-wonderful-ground-0c34d501e.yml
+++ b/.github/workflows/azure-static-web-apps-wonderful-ground-0c34d501e.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - staging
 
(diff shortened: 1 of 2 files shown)


aaronhans (approved):
cool chaining of git actions

great way to set this up to ensure staging charts are ready for review

PR #5766, opened 2021-12-10
new workflow to deploy to staging aws

aaronhans:
This is a new workflow to deploy from staging branch to AWS staging S3 bucket

This has already run successfully in the staging branch, deploying to production branch also so it doesn't get blown away somehow

### DIFF
--- a/.github/workflows/staging_aws.yml
+++ b/.github/workflows/staging_aws.yml
@@ -1,13 +1,24 @@
-name: Staging deploy -AWS
+name: Staging deploy - AWS
 on:
   push:
     branches:
       - staging
+
+# Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


jbum:
Ooh nice!

Gonna approve this, but two suggestions:

Add these two additional target trigger for manual runs and making it callable

```
  # Allows you to run this workflow manually from the Actions tab
  workflow_dispatch:
  # Allows you to call this workflow from another workflow -- see equity_deploy_staging.yml
  workflow_call:
```

2) Modify the checkout to force staging, even if run from a non-staging branch
```
      - uses: actions/checkout@v2
        with:
          ref: staging
          submodules: true
```

jbum (approved):
Approved!  See my note ^^^

aaronhans:
Thanks Jim, changes incorporated

PR #5797, opened 2021-12-16
Switched to zero uncertainty cutoff for testing sparkline chart.  Supporting forced zero cutoffs.

jbum:
Switched to using -1 to indicate no override, so that zero is a valid cutoff override.
Using a zero cutoff for testing chart, based on prior discussion with Xian and CDPH.


aaronhans (approved):
changing uncertainty_days to -1 from 0

PR #5798, opened 2021-12-16
add longer cache length to static files on covid

aaronhans:


jbum (approved):
Nice performance tweak on some files that don’t change frequently.

PR #5806, opened 2021-12-17
Dashboard date range picker as selects [To be published 12/30]

zakiya:
@jbum This work is ready for review. I deleted the tab-related code in [this commit](https://github.com/cagov/covid19/pull/5806/commits/6658d992b01544e602018f446c428ad57ff81800)

Please review to make sure I didn't bork any other site charts. 

[This commit](https://github.com/cagov/covid19/pull/5806/commits/f428e7caa2a263b99bb756d009c26b3db538f119) has the raw markup changes I made. These will need to be made in WordPress or else the padding goes wonky. 

Interactive preview at 
https://zk-dashboard-date-range-picker.pr.covid19.ca.gov/pr/zk-dashboard-date-range-picker/state-dashboard/


jbum (approved):
Just looked at the preview site.  Looking great so far Zakiya!

zakiya:
1) Padding added below chart title on smaller screens
2) Wrapped titles will have a the same height aligned bottom when the chart have a 2 up layout. 
3) Each chart filter will respect its sibling when the value changes.
4) Chart filters without siblings  will still work as expected. 

Ideally there would be a label for screen readers on each select but I struggle with what to call the first one. "Select by *type*" and "Select by *time range*" are the best I could come up with.

PR #5818, opened 2021-12-21
remove plasma and telehealth js since both of them have been retired

aaronhans:
Both plasma and telehealth pages have been retired.

This PR removes the js from the repo and removes the references to it in the js bundling code which was still including it


jbum (approved):
TY

PR #5933, opened 2022-01-13
More accurate update-dates on data-driven pages.

jbum:
Prior to this fix, the update-date at the top of state-dashboard, equity, and vaccination-progress-data reflected the data-publish date (and not any wordpress update date, even if more recent).  The date at the top of the vaccines page showed the wordpress update date only.

This fix patches all four pages to show the _later_ of the two dates: data-publish-date and wordpress-publish-date.

Normally, with no editing going on, the date will reflect the data-publish date. But if a mid-day or mid-week wordpress edit happens, the date will reflect that (until the next pipeline data update).


aaronhans (approved):
cool

PR #6014, opened 2022-01-28
we do not want the vaccines survey up anymore per Clara

aaronhans:
- Remove the survey js code from the list of imports in the bundle
- Remove the custom element for the vaccine survey from the templates


chachasikes (approved):
removes vaccine survey markup and code

jbum:
Thanks Aaron!

PR #6026, opened 2022-01-31
Changed source of cumulative vaccine doses and date.

jbum:
Changing cumulative vaccine doses (on tracker boxes) and associated data collection date, currently stored in data/daily-stats-v2.json to come from the file data/dashboard/vaccines/sparkline.json, where we store other vax related data.

This enables us to more precisely control data collection dates for Vaccine data, which is now being paused on weekends.


aaronhans (approved):
cool

PR #6069, opened 2022-02-04
Translation - state-dashboard

jorgv:


carterm (changes requested):
Line feeds are being encoded into `&#xa;` - HTML Encoded Line Feed character (using the hexadecimal value).  We need to have them remain normal line feeds.

carterm:
obsolete.  Replaced with...
https://github.com/cagov/covid19/pull/6098

PR #6074, opened 2022-02-04
PREVIEW — 14-day average for ICU beds.

jbum:
-- APPROVED FOR PRODUCTION --


chachasikes (approved):
More with the averages - and addition available ICU beds and changes over time

PR #6099, opened 2022-02-08
Dropped width: 100vw from wp-block-separator

jbum:
This removes an unsightly gap from Arabic (or any RTL language) pages. 

No apparent effect observed on LTR pages.

Thanks to @carterm for catching this and identifying the fix!

### DIFF
--- a/src/css/_general.scss
+++ b/src/css/_general.scss
@@ -1602,7 +1602,7 @@ li.active ul.nav.bd-sidenav {
 /* Wordpress Separator */
 
 .wp-block-separator {
-  border-top: 1px solid #000;
+  border-top: 1px solid $black;
   margin: 5rem calc(50% - 50vw);
 }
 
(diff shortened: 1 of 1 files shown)


carterm (approved):
removing a width adjustment

carterm (changes requested):
The HR won't extend to the right side without this.

src/css/_general.scss
```diff
@@ -1600,13 +1600,12 @@ li.active ul.nav.bd-sidenav {
 }
 
 /* Wordpress Separator */
+
 .wp-block-separator {
-  width: 100vw;
-  margin-left: calc(50% - 50vw);
-  border-top: 1px solid $black;
```
carterm, line 1606: leave the `$black` variable

carterm (approved):
should work in all cases now.  noice.

PR #6101, opened 2022-02-08
use new accordion package with our own overrides

aaronhans:
This adds new accordion to covid19
- Markup for all accordions has been updated to use the latest HTML
- Custom CSS to provide the covid19 specific accordion styles has been rewritten
- Testing in all browsers required some updates to the latest accordion, those have been incorporated in beta versions and then into new v2.0
- The latest accordion package has been installed


jbum:
I assume you're working on this...
image

aaronhans:
I fixed the chart drawer issues

zakiya (approved):
Looks okay to me but would want @jbum's review as well.

jbum:
I'm noticing that in the preview, the first accordion on the state-dashboard has significantly larger font size than the accordions that follow it (and also larger than the live site).  Some screen snapshots follow.

First accordion
CleanShot 2022-03-03 at 13 50 57@2x


Second accordion
CleanShot 2022-03-03 at 13 51 23@2x

Live site
CleanShot 2022-03-03 at 13 51 38@2x

aaronhans:
Thanks @jbum the accordion isn't doing any text size control. In the cases on this page there is an internal element in those other chart drawers that uses small-text on it.

Looks like I should also apply that to the top one too.

aaronhans:
@jbum I added the small-text class into the content of the first chart-drawer on the state dashboard wordpress post. I published that version to staging and merged just that post into this branch

Now the first chart drawer style accordion has the expected small text inside it too

jbum (approved):
Thank you!

aaronhans:
Reviewed in staging, found height bug on mobile, fixed in component.

Let's not merge until I get design review of final component, then we can install latest published version of component here

aaronhans:
This is ready to go now @jbum

- It has latest accordion package installed with multi line title expansion height bugfix
- It includes updates to accordion markup in js of last 2 chart on equity page

kkroon:
When I looked at the dashboard page in Firefox, I noticed that the "icons" are not centered:

closed

open

¯\\_(ツ)_/¯

aaronhans:
Peggy identified an issue, production blocks of accordions don't duplicate the top and bottom borders, see https://cadotgov.slack.com/archives/CVCSH3CR5/p1646936813057879

I'll look into fixing this before we merge

aaronhans:
Thank you @kkroon and @thepegisin css fixes are in for the FF misalignment and to use shared borders

PR #6107, opened 2022-02-09
PREVIEW — Aligning averages on left/right axes for cases, deaths, total-tests.

jbum:
-- APPROVED FOR PRODUCTION --

Before this patch, the left-right axes on several charts operated independently and were not aligned. This computes the average for the data on each side (bars and lines) and scales the line-axis so that the averages align.

BEFORE
image

AFTER
image


chachasikes (approved):
I see you are aligning averages now
Not sure if you want more review than an eyeball - if so, let us know

jbum:
Thanks Chach!

PR #6179, opened 2022-02-17
Including OTHER in variants chart.

jbum:
Prior to this patch, the Other category was not displayed in the Variants chart (and data was renormalized to fix it).

Due to the coming omicron-mutation, there is renewed interest in displaying this field. So the JSON options have been changed to include it again, and turn off the re-normalization feature.

Intent is to merge this in Thursday Feb 17 at 9:20 to coincide with data rollout.


aaronhans (approved):
looks good

PR #6231, opened 2022-03-02
-- Preview Only -- Homepage Rewrite

jbum:
-- PREVIEW ONLY, DO NOT MERGE --

This has a copy of the homepage at /homepage-v2/ which duplicates the content on the homepage, but in a manner more closely resembling our other pages.  Several strings which currently stored in homepage-labels are directly encoded in the page (and it's translations).

Work is still needed to figure out how to serve this content at the root URL, and handle translation navigation.


zakiya (approved):
The code and the preview look  fine to me. Not sure how I would test the WordPress of it all.

jbum:
Clsoing until I figure out a way to make this work with our build.

PR #6254, opened 2022-03-04
Aria / accessibility fixes for State Dashboard

jbum:
This reduces the valid Axe DevTools issues for the State Dashboard from 225 to 0.  Using the Axe-DevTools plugin, the page currently scores as follows, with all remaining issues incorrectly reported color-contrast.

CleanShot 2022-03-04 at 15 16 09@2x

In addition, the histogram charts have been provided with aria-labels, and are set to role="image", which prevents navigation within them, which is not useful. This fixes the read-order issues reported in the CDT accessibility report.

The following changes have been made:

* Fixed duplicate IDs.
* Removed incorrect aria-tags on pulldown menus.
* Removed aria-labels from barchart rects (that are not valid for rects).
* Added aria-labels to select elements.
* Unique label on mobile social media link.
* Histogram chart containers have an aria-label and set to role="image"


aaronhans (approved):
Impressive work Jim!

aaronhans:
This is awesome progress. 

Looks like there are more optimizations we can pursue in other PRs. Right now when I go through the state dashboard in a screen reader it is reading "unlabeled image" when it reaches a chart element. This is an improvement of it reading one of the labels out of context first though

PR #6262, opened 2022-03-07
Added daylight savings adjustments. 

jbum:
To be merged Saturday afternoon (along with similar PRs for Crons and covid-static, and covid-static-data).


aaronhans (approved):
thanks Jim

PR #6271, opened 2022-03-08
Aria fixes for Equity and Variants page -- IN PROGRESS --

jbum:
This addresses all addressable AxeDev issues, improving the site score.  

However, work continues on making the bar charts accessible.  I need to find a better method of providing narrated captions for the bars -- the method we were using (aria-label on rects) is not valid.

* Removes invalid aria-labels from rects on bar charts.
* Removes invalid roles from pull-down menus.


aaronhans (approved):
cool

PR #6304, opened 2022-03-14
Removed dodonot.njk and updated pubData.js

jbum:
These changes will enable us to remove several old unused pages (whats-open, and all the do-donot pages) without causing a build error.


aaronhans (approved):
Excellent

PR #6315, opened 2022-03-15
install this amp package so es5 build works

aaronhans:
The build was broken complaining about this package while trying to do the transpiled version. This was required for multiple packages so I installed it locally and errors went away.

Long term solution is to stop transpiling and make sure the site is still navigable without javascript on any device.

Short term fix is to obey the transpiler and deliver tons of additional dependencies to the rarer, least powerful user agents

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "covid19",
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


chachasikes (approved):
Interesting - good to know

jbum (approved):
Adds remapping package. Unsure why this is fixes it, but cool?

chachasikes:
Also @aaronhans, highlighting your important point:

**Long term solution is to stop transpiling and make sure the site is still navigable without javascript on any device.**

chachasikes:
Since we aren't very familar with what's needed (the commit only adds a package) - I'm curious about the preview feature & what packages it needs.

Dunno if we need to split dev dependencies from compilation dependencies, too - think it's fine to use utility packages as needed. Agree that our component code needs to be super simple - and anything that isn't super simple is probably a service & could be categorized as such.

PR #6388, opened 2022-03-28
fix accordions on search and update event tracking

aaronhans:
This fixes some bugs introduced with the accordion upgrades;

- Accordions used for quick answers in search results started defaulting to open. This is the failure state and is caused by a js error looking for new markup while search results pages uses old accordions. This is a friendly way to fail but not desired because they take up tons of space above basic search results.
- Accordion open tracking code wasn't updated to look for new selector pattern
- Automated E2E tests need to use new accordion pattern


jbum (approved):
Thanks for fixing this Aaron!

PR #6476, opened 2022-04-14
Skip localization code block for fragments.

jbum:
This prevents a build error when PDF links containing "--en.pdf" are used in HTML fragments.


aaronhans (approved):
nice

PR #6500, opened 2022-04-18
Increased hover area on tooltips.

jbum:
Currently, on the state-dashboard, tool-tips appear when you hover certain parts of the histogram charts (cases, deaths, tests, hospitalizations).  These tool-tips show specific values for the data in the charts for the date you are hovering over.  Like so:

CleanShot 2022-04-18 at 15 00 01

At the moment, the site is programmed so that these tool-tips only appear when you mouse over specific graphic elements (bars and lines).  

For some of the charts, this is too restrictive, especially when the bars and lines are low.  This patch deletes several lines of code and modifies the "hover area" to include the entire chart area.   This makes it easier to discover and use the tool-tip feature.

The change in the "hover area" is as shown:

image


zakiya (approved):
Noticeable improvement on the state-dashboard charts. Nice.

PR #6514, opened 2022-04-21
Fixed y-axis scaling and tick issues for health equity chart.

jbum:
Fixes a Y-axis scaling problem, and increases tick count / formatting for these low numbers we are currently seeing.

Before:
CleanShot 2022-04-21 at 13 16 17

After:
CleanShot 2022-04-21 at 13 17 11


aaronhans (approved):
clean log

aaronhans (approved):
cool

PR #6531, opened 2022-04-27
all time top queries included

carterm:
Re-activating "Popular Searches" and including the all time top queries so they are visible all the time.  The top queries for the week will still show up at the top.


aaronhans (approved):
cool

PR #6590, opened 2022-05-18
Autocomplete patch, eliminates use of mark, and fixes mark bug.

jbum:
**NOTE: Before merging, let's test all instances of auto-complete.**

Prior to this patch, there were two issues with auto-complete on county-search.

1) A recently introduced bug (of undetermined cause) caused it fail on AWS hosted instances, returning a mark that contained angle brackets, like so:

CleanShot 2022-05-18 at 14 26 10

2) Even when working, the results looked wonky, since they inserted a space and highlighted region.

This work-around patch replaces the awesomplete.ITEM function with a simplified version which makes the results look better.

CleanShot 2022-05-18 at 14 27 03

### DIFF
--- a/src/js/common/search/index.js
+++ b/src/js/common/search/index.js
@@ -73,25 +73,35 @@ class CAGovCountySearch extends window.HTMLElement {
     rtlOverride(this, 'div', 'ltr');
   }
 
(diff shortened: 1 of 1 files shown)


aaronhans (approved):
nice fix, thanks Jim

PR #6661, opened 2022-06-15
Forcing staging build for staging workflow. 

jbum:
Oddly, our staging build's NODE_ENV environment variable has stopped registering as 'staging' or 'developemnt' -- it is coming up as 'undefined' which is causing staging-specific logic to be skipped.  A symptom of this is the staging.covid19 equity page is loading the production data.  In the below snapshot, the pathname of the json should be on the 'to-review' path, not 'reviewed'.

Evidence:
CleanShot 2022-06-15 at 10 07 38

This patch forces a staging build.  After merging, I'll apply this to the staging branch as well.

### DIFF
--- a/.github/workflows/staging_aws.yml
+++ b/.github/workflows/staging_aws.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Build 11ty
         run: |
           npm install
-          npm run build
+          npm run build:staging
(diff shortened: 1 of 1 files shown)


xjensen (approved):
Looks good to me.

PR #6742, opened 2022-07-14
fixed disparity chart tooltip

JamesSLogan:
Chris Hoover pointed out the error in slack. Seems that the value was accidentally treated as a percentage.


jbum (approved):
Thanks for catching this James -- this was almost certainly a cut & paste bit of code from the Variants chart (which displays those numbers as ratios converted to percentages).

PR #6791, opened 2022-08-03
Add probable cases to staging branch

JamesSLogan:
- Both branches contain new state-dashboard-v2.html post
- Added src/js/dashboard-v2 to support v2 state dashboard


jbum (approved):
Looks workable. Thanks!

PR #7065, opened 2022-11-28
add GA code

JamesSLogan:
Stakeholder asked for tags on every page. It does not appear to negatively impact loading times (tested using home page, comparing 2 pr sites). 2 extra requests are made each page load.


jbum (approved):
Well I don't love it, in that we already had analytics in place.  But I see it's using the newer tag manager, and I believe the old style analytics are getting phased out in a few months, so probably for the best.  It might good to tweak some of the existing reporting (see use of ga() and reportGA() function) to make use of it.

PR #7121, opened 2022-12-19
Patch for broken feedback service.

jbum:
Replaced use of api.alpha.ca.gov with fa-go-alph-d-002.azurewebsites.net (the underlying service) due to DNS config issues.


aaronhans (approved):
Thanks Jim!

PR #7256, opened 2023-02-23
update page feedback

aaronhans:
We have deployed new page feedback architecture running in AWS.

This new data store is now being consumed by the data team and powers the dashboards used by site reviewers

This update will change the covid site frontend to post to our new AWS endpoints


JamesSLogan (approved):
cool! I'll reach out via email about some details I need for documentation purposes.

aaronhans:
Thanks @JamesSLogan the page feedback component is a universal item that is part of the design system and is used on many sites so the infrastructure will continue to be maintained by ODI.

PR #7287, opened 2023-03-13
Remove 'Quick Answers' result from search

jbum:
The Azure QnA Maker service is being sunsetted.  This removes the 'Quick Answers' feature from search results, which relies on that feature.  The removal of the feature was approved by Josh Pocus.

Once this is merged, we can disable the fa-go-covid-answ-001 trigger on Azure, as well as take down our QnA Maker setup (I suggest waiting a week or two).


JamesSLogan (approved):
Looks good👍 I forgot that searching is a bit broken on PR sites but it worked on a local build.

aaronhans (approved):
Nice Jim!

PR #7328, opened 2023-04-03
decluttered variants chart

JamesSLogan:


jbum (approved):
So much easier on the eyes.  Thank you James!

PR #7605, opened 2023-08-09
Updated Sass version to fix issue with local build.

jbum:
I was seeing an error message on my arm64 Macintosh when attempting to run the site locally with a more recent version of node.  Upgrading Sass from 8 -> 9 fixed it.

image


JamesSLogan (approved):
works on my system 👍

PR #7804, opened 2024-04-19
Support use of uppercase in branch names for PR previews.

jbum:
Support use of uppercase in branch names for PR previews.

Prior to this fix, if a branch name contained any uppercase letters, the PR Preview feature would 404.


JamesSLogan (approved):
Looks good, thanks for this fix!

